PR #64, opened 2019-05-01
Handle error responses

harrietc52:


app/controllers/v1/chips_controller.rb
```diff
@@ -8,7 +8,8 @@ def update
       chip.update(attributes)
       head :ok
     rescue StandardError => e
-      render json: { errors: e.message }, status: :unprocessable_entity
+      data = { data: { errors: e.message } }
+      render json: data, status: :unprocessable_entity
```
stevieing: Do we need this extra assignment as it is fairly simple? 
render json: { data: { errors: e.message } }, status: :unprocessable_entity

harrietc52: Sure, I'll changes the ones I can without rubocop shouting

PR #103, opened 2019-07-18
Gpl 101 pipelines documentation

stevieing:


app/pipelines/pipelines/requestor/factory.rb
```diff
@@ -14,34 +17,51 @@ module Factory
...
+      # checks if the factory is valid, if so will save all of the requests
+      # @return [Boolean] if the requests have been successfully saved or not
       def save
         return false unless valid?
 
         requests.collect(&:save)
         true
       end
 
+      # Takes each set of request attributes:
```
harrietc52: These bullet points aren't quite working I don't think

app/pipelines/pipelines/requestor/factory.rb
```diff
@@ -51,6 +71,9 @@ def build_requests(attributes)
         end
       end
 
+      # Validates the requests:
```
harrietc52: And these

app/pipelines/pipelines/requestor/model.rb
```diff
@@ -3,6 +3,12 @@
 module Pipelines
   module Requestor
     # Model - behaviour for pipeline requests model
+    # includes various bits of behaviour in the model:
+    # * the Material concern
+    # * request association
+    # * sample association
+    # * delegation of sample attributes readers
+    # * validation to check that each of the request attributes is present
```
harrietc52: These are though!

app/pipelines/pipelines/requestor/factory.rb
stevieing: so are these ...

harrietc52: Ok can leave it then, happy to merge

PR #104, opened 2019-07-23
Modified message for pacbio so it is no longer a flat message.

stevieing:


spec/pipelines/pacbio/message_spec.rb
```diff
@@ -25,72 +27,98 @@
...
-    it 'must have a study_uuid' do
-      expect(key[:study_uuid]).to eq(well.library.request.external_study_id)
+    it 'must have a plate barcode' do
+      expect(key[:plate_barcode]).to eq(plate.barcode)
     end
 
-    it 'must have a id_pac_bio_run_lims' do
-      expect(key[:id_pac_bio_run_lims]).to eq(well.plate.run.id)
+    it 'must have a plate uuid' do
+      expect(key[:plate_uuid]).to eq(plate.uuid)
```
harrietc52: I think this is plate_uuid_lims

PR #107, opened 2019-07-29
Gpl 100 multiplexing

stevieing:


app/controllers/v1/pacbio/libraries_controller.rb
```diff
@@ -38,11 +38,25 @@ def library
...
+        params.permit(:volume, :concentration, :library_kit_barcode, :fragment_size, :relationships)
+              .to_h.tap do |library|
+          library[:requests] = request_param_names(params)
         end
       end
 
+      def request_param_names(params)
+        params.require(:relationships)[:requests].require(:data).map do |request|
+          request.permit(:id, :type).to_h.tap do |tag|
+            tag[:tag] = request.require(:relationships)[:tag].require(:data).permit(:id, :type).to_h
```
harrietc52: Maybe we could pull `request.require(:relationships)[:tag].require(:data).permit(:id, :type).to_h`  out to `tag_params_names` method for consistency and readability

spec/requests/v1/pacbio/libraries_spec.rb
```diff
@@ -71,184 +89,225 @@
...
+                                  type: 'tags',
+                                  id: tag.id
+                                }
+                              }
+                            }
+                          }
+                        ]
+                      }
+                    }
+                },
```
harrietc52: Should we add a seperate test for when a library is created with two requests

spec/requests/v1/pacbio/libraries_spec.rb
```diff
@@ -23,38 +27,52 @@
...
 
       expect(json['data'][1]['attributes']['volume']).to eq(library2.volume)
       expect(json['data'][1]['attributes']['concentration']).to eq(library2.concentration)
       expect(json['data'][1]['attributes']['library_kit_barcode']).to eq(library2.library_kit_barcode)
       expect(json['data'][1]['attributes']['fragment_size']).to eq(library2.fragment_size)
-      expect(json['data'][1]["attributes"]["pacbio_tag_id"]).to eq(library2.pacbio_tag_id)
-      expect(json['data'][1]["attributes"]["pacbio_request_id"]).to eq(library2.pacbio_request_id)
     end
 
   end
```
harrietc52: we could add a test for update too

spec/models/pacbio/library_spec.rb
```diff
@@ -18,12 +18,29 @@
...
+
+    it 'can have one or more' do
+      library = create(:pacbio_library)
+      (1..5).each do |i|
+        create(:pacbio_request_library, request: create(:pacbio_request), library: library, tag: create(:tag))
+      end
+      expect(library.requests.count).to eq(5)
+    end
+  end
+
```
harrietc52: Could also add tests for `well_libraries` and `request_libraries`

spec/models/pacbio/request_spec.rb
```diff
@@ -4,4 +4,14 @@
...
+  context 'libraries' do
+    it 'can have one or more' do
+      request = create(:pacbio_request)
+      (1..5).each do |i|
+        create(:pacbio_request_library, request: request, library: create(:pacbio_library), tag: create(:tag))
+      end
+      expect(request.libraries.count).to eq(5)
+    end
+  end
+
```
harrietc52: could add test also for `request_libraries`

spec/models/pacbio/request_library_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  it 'can have a sample name' do
+    expect(create(:pacbio_request_library).sample_name).to be_present
+  end
+
+  it 'can have some tag attributes' do
+    request_library = create(:pacbio_request_library)
+    expect(request_library.tag_oligo).to be_present
+    expect(request_library.tag_group_id).to be_present
+    expect(request_library.tag_set_name).to be_present
+  end
```
harrietc52: nice :) `tag_id` too

spec/models/pacbio/run_spec.rb
```diff
@@ -37,18 +37,18 @@
   end
```
harrietc52: maybe a test for `plate_wells`

spec/models/pacbio/well_spec.rb
```diff
@@ -84,4 +88,37 @@
...
+      expect(sample_names.length).to eq(2)
+      expect(sample_names.first).to eq(request_libraries.first.request.sample_name)
+    end
+
+    it 'can return a list of tags' do
+      expect(well.tags).to eq(request_libraries.collect(&:tag_id))
+    end
+
+  end
+
```
harrietc52: tests for `sample_names` and `tags` methods

app/resources/v1/pacbio/request_library_resource.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module V1
+  module Pacbio
+    # LibraryResource
+    class RequestLibraryResource < JSONAPI::Resource
+      model_name 'Pacbio::RequestLibrary'
+
+      attributes :sample_name, :tag_oligo, :tag_group_id, :tag_set_name
```
harrietc52: do you want `tag_id` too?

config/pipelines.yml
```diff
@@ -67,17 +67,17 @@ default: &default
...
-              value: library
+              value: request_libraries
               children:
                 cost_code:
                   type: :model
                   value: request.cost_code
                 pac_bio_library_tube_id_lims:
                   type: :model
-                  value: id
+                  value: library.id
```
harrietc52: thinking this would need to be `libraries`?

config/pipelines.yml
```diff
@@ -67,17 +67,17 @@ default: &default
...
                 pac_bio_library_tube_uuid:
                   type: :model
-                  value: uuid
+                  value: library.uuid
```
harrietc52: and this

config/pipelines.yml
```diff
@@ -97,8 +97,8 @@ default: &default
                   type: :model
                   value: tag.id
                 tag_set_name:
-                  type: :string
-                  value: pacbio
+                  type: :model
+                  value: tag.set_name
```
harrietc52: and this `tags` possibly

config/pipelines.yml
stevieing: It needs to be request libraries. Libraries on their own don't contain sufficient information such as sample name and tag information. If you look at the way libraries are created we create the library and then add all of the requests. A library can contain many requests with tags and each message needs an associated tag. Hopefully that makes sense?

spec/models/pacbio/library_spec.rb
stevieing: I did think about that but the well_libraries is purely a join table and I thought the fact that you can add libraries to a well proves that well_libraries are working correctly. What do you think?
With request_libraries again I did think about that. I have added tests below and the fact that a library has requests proves it is working correctly. Happy to add them if you think it is necessary.

db/schema.rb
```diff
@@ -35,6 +35,15 @@
     t.index ["pacbio_run_id"], name: "index_pacbio_plates_on_pacbio_run_id"
```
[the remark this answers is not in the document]
stevieing: I thought I had already done this. Turns out the migration was wrong and I was removing the wrong references. Good spot! done.

db/schema.rb
```diff
@@ -152,6 +161,14 @@
     t.datetime "updated_at", null: false
```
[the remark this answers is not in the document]
stevieing: As above. Migration was wrong. Done.

app/controllers/v1/pacbio/libraries_controller.rb
stevieing: agreed. Done. It might be a good idea to abstract the behaviour out for params names as there is quite a lot of it in the controllers. I will add it as an issue.

spec/requests/v1/pacbio/libraries_spec.rb
stevieing: Good spot :) I did have multiple requests wrapped up in another test but I wasn't checking whether those requests existed. Turns out the associations weren't being saved so I have set autosave to true. I am happy to leave it like that. Do you think it needs to be in its own test?

spec/models/pacbio/request_library_spec.rb
stevieing: done.

spec/models/pacbio/run_spec.rb
stevieing: done.

spec/models/pacbio/well_spec.rb
stevieing: The tests are there under the request_libraries context. Is that what you are asking?

spec/requests/v1/pacbio/libraries_spec.rb
stevieing: Agreed. As discussed we will add this as an enhancement issue.

spec/requests/v1/pacbio/libraries_spec.rb
harrietc52: If we can just test in the existing test that the multiple requests exist, that is maybe enough? :)

spec/models/pacbio/well_spec.rb
harrietc52: Ah yep sorry! Missed those

config/pipelines.yml
harrietc52: So this value should be changed to `request_libraries`? Would you loop throgh their id's?

config/pipelines.yml
harrietc52: Same with this, if it is tags, do you loop through each tags set_name, or is this covered with the nested message you mentioned?

spec/models/pacbio/library_spec.rb
harrietc52: Think you're right and `libraries` proves the join is working as expected, same with request_libraries

spec/requests/v1/pacbio/libraries_spec.rb
stevieing: Yes. I have added an extra expectation the multiple libraries test which checks that two requests exist.

PR #122, opened 2019-08-12
Added some documentation mostly for pacbio and a bit to the message c…

stevieing:
…reation.

### DIFF
--- a/app/messages/messages/message.rb
+++ b/app/messages/messages/message.rb
@@ -2,11 +2,20 @@
 
 module Messages
   # Message
+  # Creates a message in the correct structure for the warehouse
   class Message
     include ActiveModel::Model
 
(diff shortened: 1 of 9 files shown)


app/messages/messages/message.rb
```diff
@@ -27,8 +42,16 @@ def build_children(object, field)
       end
     end
 
-    private
```
pjvv: Was this intentional?

stevieing: Yes. I wanted to add some comments to the method and they don't show if the method is private. The method also has some duplication with some work Harrys has done so we will probably abstract the method out so it will need to be public anyway.

pjvv: OK great thanks.

PR #147, opened 2019-10-11
Added libraries relationship to saphyr requests. Created rake task fo…

stevieing:
…r adding historical Saphyr runs.


lib/tasks/migrate_saphyr_data.rake
```diff
@@ -0,0 +1,48 @@
...
+
+require 'csv'
+
+namespace :migrate_saphyr_data do
+  task create_runs: :environment do |_t|
+    # samples and requests
+    table = CSV.parse(File.read(Rails.root.join('lib', 'data', 'saphyr_samples.csv')), headers: true)
+    table.each do |row|
+      created_at = Date.parse(row['created_at'])
+      sample = Sample.create!(name: row['name'], species: row['species'], external_id: row['external_id'], created_at: created_at)
```
harrietc52: This samples table has a external_study_id column, maybe this can be removed if the external_study_id is on the request?

app/models/saphyr/request.rb
```diff
@@ -9,5 +9,8 @@ module Saphyr
   # A saphyr request can have one sample
   class Request < ApplicationRecord
     include Pipelines::Requestor::Model
+
+    has_many :libraries, class_name: 'Saphyr::Library', foreign_key: :saphyr_request_id,
```
harrietc52: Could maybe add a unit test for this

PR #151, opened 2019-10-16
Backload Saphyr Data

stevieing:


config/pipelines.yml
```diff
@@ -146,3 +146,6 @@ development:
 
 test:
   <<: *default
+
+staging:
+  <<: *default
```
harrietc52: Do we need to add production too?

PR #154, opened 2019-10-18
Added the ability to add a request to pacbio library with no tags.

stevieing:
Because we are not doing multiplexing at the moment I have not added checks to ensure that all of the requests in a library have tags if there is more than one request.

### DIFF
--- a/spec/models/pacbio/well_library_factory_spec.rb
+++ b/spec/models/pacbio/well_library_factory_spec.rb
@@ -43,6 +43,8 @@ RSpec.describe Pacbio::WellLibraryFactory, type: :model, pacbio: true do
       expect(factory).to_not be_valid
       expect(factory.errors.full_messages).to_not be_empty
     end
+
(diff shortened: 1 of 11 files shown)


spec/models/pacbio/well_library_factory_spec.rb
```diff
@@ -43,6 +43,8 @@
       expect(factory).to_not be_valid
       expect(factory.errors.full_messages).to_not be_empty
     end
+
+    it 'produces an error if there are multiples requests in any of the libraries and they do not have tags'
```
harrietc52: Think the body of this test is missing?

stevieing: This was deliberate. I need to add a feature to library factory and well library factory to check for empty tags when it is multiplexed and didn't know the best way to do it. As it was not required for first pacbio release I left it. I will try and add something to the test to indicate that ...

PR #206, opened 2020-01-07
Fix removing libraries from a well

harrietc52:
Added lots of tests
Removed validation that libraries have to exist,
as libraries might be an empty list

### DIFF
--- a/spec/requests/v1/pacbio/wells_spec.rb
+++ b/spec/requests/v1/pacbio/wells_spec.rb
@@ -171,10 +171,8 @@ RSpec.describe 'WellsController', type: :request do
 
           expect(errors).to include('plate')
           expect(errors).to include('movie_time')
-          expect(errors).to include('libraries')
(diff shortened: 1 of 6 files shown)


harrietc52:
Failing on travis but not locally. I think it is to do with the way tags are dynamically created, and they have to be unique?!

spec/requests/v1/pacbio/wells_spec.rb
```diff
@@ -324,7 +324,7 @@
     context 'when successfully adding a new library' do
       let(:uniq_tag) { create(:tag, set_name: 'pipeline2') }
       let(:request_library1) { create(:pacbio_request_library, tag: uniq_tag) }
-      let(:updated_libraries_data) { existing_libraries_data.push({ type: "libraries", id: request_library1.id }) }
+      let(:updated_libraries_data) { existing_libraries_data.push({ type: "libraries", id: request_library1.library.id }) }
```
harrietc52: Ah! Thank you!

PR #237, opened 2020-01-20
Added validation for cost code on library creation

BenTopping:
Added tests

### DIFF
--- a/spec/models/pacbio/library_factory_spec.rb
+++ b/spec/models/pacbio/library_factory_spec.rb
@@ -30,29 +30,30 @@ RSpec.describe Pacbio::LibraryFactory, type: :model, pacbio: true do
       expect(factory.libraries[0].tube).to be_present
       expect(factory.libraries[0].request_libraries.length).to eq(2)
     end
+  end
 
+  context '#validation' do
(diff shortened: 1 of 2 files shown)


spec/models/pacbio/library_factory_spec.rb
```diff
@@ -42,6 +44,14 @@
       expect(factory).to_not be_valid
       expect(factory.errors.full_messages).to_not be_empty
     end
+
+    it 'produces an error if the request contains an empty cost_code' do
+      library_attributes << attributes_for(:pacbio_library).merge(requests: [{id: request_empty_cost_code.id, type: 'requests'}])
+      factory = Pacbio::LibraryFactory.new(library_attributes)
+      expect(factory).to_not be_valid
+      expect(factory.errors.full_messages).to eq(['Cost code must be present'])
+    end
```
KatyTaylor: Hi @BenTopping , Harry and I were looking at this together and we reckon it would be nicer to call 'save' explicitly before the expectations. We realised that the only reason it's calling the validation in here is because the 'be_valid' is calling it, which seems a bit crazy and magical. Does that make sense? Thanks

PR #256, opened 2020-02-03
Gpl 266 sample sheet fields

KatyTaylor:
- add new fields into sample sheet (generated from Traction, given to SMRT link software to tell it about what's being sequenced)
- restructure sample sheet to match new specification - have multiplexed samples with their tags (barcodes) on secondary rows under a header row for the well
- add TagSet model as parent to Tag (removed Set Name from Tag model as part of this change) - will need to run a script to insert a TagSet and update the Tags:

`set = TagSet.create!(name: 'Sequel_16_barcodes_v3', uuid: '4d87a8ab-4d16-f0b0-77e5-0f467dba442e')`

`Tag.update_all(tag_set_id: set.id)`


spec/factories/pacbio/wells.rb
```diff
@@ -20,5 +20,11 @@
         well.libraries = create_list(:pacbio_request_library_with_tag, 5).collect(&:library)
       end
     end
+
+    factory :pacbio_well_with_request_libraries_no_tag do
+      after(:create) do |well|
+        well.libraries = create_list(:pacbio_request_library, 5).collect(&:library)
```
harrietc52: Is the `&` required?

spec/csv_generator/csv_generator_spec.rb
```diff
@@ -29,7 +29,142 @@
       expect(headers).to eq(expected_headers)
     end
```
harrietc52: Nothing to do with your changes but think there is an `expect(true).to be_truthy` above! Seems a bit redundant?

spec/models/pacbio/request_library_spec.rb
```diff
@@ -22,7 +22,7 @@
     request_library = create(:pacbio_request_library, tag: create(:tag))
     expect(request_library.tag_oligo).to be_present
     expect(request_library.tag_group_id).to be_present
-    expect(request_library.tag_set_name).to be_present
+    expect(request_library.tag.tag_set.name).to be_present
```
harrietc52: Does this want to be `tag_set_name` to test the delegate method? Maybe `expect(request_library.tag_set_name).to eq request_library.tag.tag_set.name`

spec/models/pacbio/well_spec.rb
```diff
@@ -130,4 +124,12 @@
 
   end
 
+  context 'sample sheet mixin' do
```
harrietc52: 👍

spec/models/tag_spec.rb
```diff
@@ -14,15 +14,15 @@
...
+
   it 'group id should be unique within set' do
     tag = create(:tag)
     expect(build(:tag, group_id: tag.group_id)).to_not be_valid
   end
 
-  it 'is not valid without a set name' do
-    expect(build(:tag, set_name: nil)).to_not be_valid
-  end
-
```
harrietc52: Could test the delegate name here possibly

spec/pipelines/pacbio/message_spec.rb
```diff
@@ -112,7 +112,7 @@
           end
 
           it 'can have a tag set name' do
-            expect(message_sample[:tag_set_name]).to eq(request_library.tag.set_name)
+            expect(message_sample[:tag_set_name]).to eq(request_library.tag.tag_set.name)
           end
```
harrietc52: Doesn't really matter but can use the `tag_set_name` on request libraries again 😊

spec/requests/v1/pacbio/wells_spec.rb
```diff
@@ -322,7 +322,7 @@
     end
 
     context 'when successfully adding a new library' do
-      let(:uniq_tag) { create(:tag, set_name: 'pipeline2') }
+      let(:uniq_tag) { create(:tag, tag_set: TagSet.find_or_create_by(name: 'Test Tag Set 2', uuid: '1234567')) }
```
harrietc52: Could use factory bot `let(:tag_set) { create(:tag_set) }`, then  `let(:uniq_tag) { create(:tag, tag_set: tag_set)}`

config/routes.rb
```diff
@@ -6,6 +6,7 @@
   namespace :v1 do
     jsonapi_resources :samples, only: %i[index create]
     jsonapi_resources :tags,    only: %i[index create update destroy]
+    jsonapi_resources :tag_sets,    only: %i[index create update destroy]
```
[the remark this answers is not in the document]
harrietc52: Ahh sorry just seen below you've added supporting tests for these 👍

app/models/pacbio/well.rb
```diff
@@ -25,10 +26,6 @@ def position
       "#{row}#{column}"
     end
 
-    def position_leading_zero
-      "#{row}#{column.rjust(2, '0')}"
-    end
-
     def summary
       "#{sample_names},#{comment}"
```
harrietc52: This is the methods that gives us the trailing `,` in run comments. We could add that as an issue to fix another time? 🙂

KatyTaylor: 👍 #258

db/migrate/20200123163055_create_tag_sets.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTagSets < ActiveRecord::Migration[5.2]
+  def change
+    create_table :tag_sets do |t|
+      t.string :name
+      t.string :uuid
+      t.timestamps
+    end
+  end
```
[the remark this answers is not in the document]
KatyTaylor: I only added one for Tags because I tried to reverse it locally and rails complained. It didn't complain about Tag Sets so think I can let the default do its thing?

KatyTaylor:
> All really great - few minor ideas, can choose the ones you agree with!

Thanks Harry - I've made the changes and will merge it in when it's built.

PR #257, opened 2020-02-03
Added pac_bio_run_name field to mlwh message for pacbio run

harrietc52:


harrietc52:
On deployment, run script to copy the values from id_pac_bio_run_lims into the new pac_bio_run_name column? Confirm what data is needing to be updated

Something like:
`UPDATE pac_bio_run SET pac_bio_run_name= id_pac_bio_run_lims`

But check constraints/ null checks etc

PR #262, opened 2020-02-04
Changed date format to have 24 hour time (opposed to 12 hour)

BenTopping:
debugging line removed from test

### DIFF
--- a/config/initializers/date_and_time_formats.rb
+++ b/config/initializers/date_and_time_formats.rb
@@ -2,4 +2,4 @@
 
 # Be sure to restart your server when you modify this file.
 
-Time::DATE_FORMATS[:us] = '%Y/%m/%d %I:%M'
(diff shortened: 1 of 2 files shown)


config/initializers/date_and_time_formats.rb
```diff
@@ -2,4 +2,4 @@
 
 # Be sure to restart your server when you modify this file.
 
-Time::DATE_FORMATS[:us] = '%Y/%m/%d %I:%M'
+Time::DATE_FORMATS[:us] = '%Y/%m/%d %H:%M'
```
KatyTaylor: While you're at it, I think d/m/y is more readable - what do you think? Maybe it's not obvious unless you're british though?

BenTopping: At the moment time needs to be that structure so that the tables in traction sort correctly by created_at because bootstrap vue does not sort dates. I think its under issue 290 in traction ui issues 👍

PR #287, opened 2020-02-27
tag_set_id_lims is now tag id and tag_identifier is tag group_id.

stevieing:


config/pipelines.yml
```diff
@@ -95,10 +95,10 @@ default: &default
                   value: tag.oligo
                 tag_set_id_lims:
                   type: :model
-                  value: tag.group_id
+                  value: tag.id
```
KatyTaylor: I think it should be tag.tag_set.id

PR #298, opened 2020-03-05
Gpl 320 upgrade to rails 6

stevieing:
There is some stuff that I have had to do that might need some explanation so happy to do a group code review if you both feel that would be beneficial.
Based on what I have had to do to get this to work that it would be beneficial to add a new story to persist the configuration.


bin/rails
```diff
@@ -1,11 +1,4 @@
 #!/usr/bin/env ruby
-# frozen_string_literal: true
-
-begin
-  load File.expand_path('spring', __dir__)
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
```
KatyTaylor: What's the reason for removing these bits, and should you also remove bin/spring ?

config/cable.yml
```diff
@@ -2,9 +2,9 @@ development:
   adapter: async
 
 test:
-  adapter: async
+  adapter: test
```
KatyTaylor: what's the reason for this change?

config/cable.yml
```diff
@@ -2,9 +2,9 @@ development:
...
 
 production:
   adapter: redis
   url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
-  channel_prefix: traction-service_production
+  channel_prefix: traction_service_production
```
KatyTaylor: also, what will this change affect?

config/initializers/cors.rb
```diff
@@ -1,18 +1,17 @@
...
-end
+# Rails.application.config.middleware.insert_before 0, Rack::Cors do
+#   allow do
+#     origins 'example.com'
+#
+#     resource '*',
+#       headers: :any,
+#       methods: [:get, :post, :put, :patch, :delete, :options, :head]
+#   end
+# end
```
KatyTaylor: How is CORS handled now?

spec/models/concerns/material_spec.rb
```diff
@@ -1,5 +0,0 @@
-require 'rails_helper'
-
-shared_examples "material" do
-  it { is_expected.to have_one(:tube) }
-end
```
KatyTaylor: why get rid of this test?

bin/rails
stevieing: Part of the Rails upgrade. A number of files are replaced automatically during the upgrade so these were over written.

config/cable.yml
stevieing: As above.

config/cable.yml
stevieing: As above. We don't use action cable.

config/initializers/cors.rb
stevieing: As above. We handle cors in application.rb. We could move it to initializer but not a problem.

spec/models/concerns/material_spec.rb
stevieing: This is a shared example and is a duplicate of spec/support/material.rb. I was getting a redefinition warning.

PR #299, opened 2020-03-06
Automated semantic releases

stevieing:


pjvv:
Looks good, should try it with a new line in the `.release-version` file to confirm all works as expected. I think this repo should start using the same branching strategy (only develop and uat) as most of the other repos in sanger to minimise confusion and errors if we copy paste config from it.

PR #329, opened 2020-04-22
Gpl 385 install graphql

sdjm-tessella:

* Install GraphQL.
* Add an example query for getting Wells.
* Add an example query for mutating Well positions.


app/graphql/mutations/update_well_position.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Mutations
+  # Mutation to handle the update of a well position for a well with a given ID.
+  class UpdateWellPosition < BaseMutation
```
Chris-Friend, line 5: Would this be better named something like `UpdateWellPositionMutation` to more easily identify it as a mutation?

app/graphql/types/well_query_type.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Types
+  # The type for Well queries.
+  class WellQueryType < BaseObject
```
Chris-Friend, line 5: Does it make sense to add this to a Types::Queries namespace structure?

app/graphql/types/well_mutation_type.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Types
+  # The type for Well mutations.
+  class WellMutationType < BaseObject
```
Chris-Friend, line 5: Similary does it make sense to add this to a Types::Migrations namespace structure?

app/graphql/types/well_query_type.rb
sdjm-tessella: We can if you think it makes it clearer.  I'm just following how GraphQL suggest to lay these classes out.  `BaseObject` is in the `Types` namespace so it might not be appropriate.

app/graphql/mutations/update_well_position.rb
sdjm-tessella: I did do that at first.  I felt that being in the `Mutations` namespace made it unnecessary to name the class with `Mutations` as well.

app/graphql/types/well_query_type.rb
Chris-Friend: :+1: I suppose this can always be changed later quite easily if we feel it needs to be

app/graphql/types/well_mutation_type.rb
Chris-Friend: Resolved as per comment on similar conversation

app/graphql/types/well_query_type.rb
```diff
@@ -0,0 +1,26 @@
...
+      Well.find(id)
+    end
+
+    field :wells, [WellType], null: false do
+      description 'Find all wells, optionally those associated with a plate.'
+      argument :plate_id, Int, required: false
+    end
+
+    def wells(plate_id: nil)
+      wells = Well.all
```
stevieing, line 21: Well.find_by(plate_id: plate_id)

stevieing: Sorry. It looks like you are trying to find multiple wells so it should be Well.where(plate_id: plate_id) I think. The docs should tell you exactly how to do this.

stevieing, line 21: I think you can use where here? e.g. Well.where(plate_id: plate_id). Not too sure on the syntax. Docs should give further info.

sdjm-tessella: Yes good idea

PR #335, opened 2020-04-24
Gpl 385 create ont entities

sdjm-tessella:
Closes #

*
*
* ...


spec/requests/v2/plates_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+        )
+        {
+          plate { barcode wells { material { ... on Request { sample { name externalId } } } } }
+          errors
+        }
+      }
+      GQL
+    end
+
+    it 'provides an error when data was invalid' do
```
Chris-Friend, line 131: Does this test the response when the request factory returns errors?
If so would it not be better mocking out the factory to return errors irrespective of the input, so you don't have factory logic leaking into these tests?
If not, does this not just test the same thing as the below `provides an error when missing required fields` test, and a new test should be added to test getting errors from the factory?

sdjm-tessella: Excellent point, and yes it is testing that we handle errors coming back from the factory.  I'll mock out the error response using the original GraphQL query from the first test.

PR #341, opened 2020-04-28
Gpl 426 tag taggable join

Chris-Friend:

* Add a new Taggable concern (currently just included by Ont::Requests)
* Add a new TagTaggable join between Tag and Taggable (since a taggable can have many tags, and a tag many taggables)
* Test the new models and updated relationships


db/migrate/20200428064825_create_tag_taggables.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTagTaggables < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tag_taggables do |t|
+      t.belongs_to :taggable, polymorphic: true, index: true
+      t.belongs_to :tag, polymorphic: true, index: true
```
sdjm-tessella, line 5: Is this supposed to be polymorphic?

db/schema.rb
```diff
@@ -193,6 +193,17 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "tag_taggables", force: :cascade do |t|
+    t.string "taggable_type"
+    t.integer "taggable_id"
+    t.string "tag_type"
+    t.integer "tag_id"
```
sdjm-tessella, line 200: Then these two columns should become just the one for the tag ID

spec/models/tag_spec.rb
```diff
@@ -32,4 +32,27 @@
...
+
+  it 'can have no tag taggables' do
+    tag = create(:tag_with_taggables, taggables_count: 0)
+    expect(tag.tag_taggables).to be_empty
+  end
+
+  it 'can have many tag taggables' do
+    numTaggables = 3
+    tag = create(:tag_with_taggables, taggables_count: 3)
+    expect(tag.tag_taggables.count).to eq(3)
```
sdjm-tessella, line 44: I'm guessing you meant to use the variable above both here and on the line above?  Either way, if you do, you need to use snake_case and if you don't the variable needs removing.

spec/models/tag_spec.rb
```diff
@@ -32,4 +32,27 @@
...
+  end
+
+  it 'on destroy destroys tag taggables, not taggables' do
+    numTaggables = 3
```
sdjm-tessella, line 48: snake_case needed here

spec/support/taggable.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  it 'returns all tags with many tag taggables' do
+    taggable = create(taggable_model)
+    expect(taggable.tags).to_not be_empty
+    expect(taggable.tags).to eq(taggable.tag_taggables.map { |tag_taggable| tag_taggable.tag })
+  end
+
+  it 'on destroy destroys tag taggables, not taggables' do
+    numTags = 3
```
sdjm-tessella, line 14: snake_case

spec/support/taggable.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(taggable.tags).to be_empty
+  end
+
+  it 'returns all tags with many tag taggables' do
+    taggable = create(taggable_model)
+    expect(taggable.tags).to_not be_empty
+    expect(taggable.tags).to eq(taggable.tag_taggables.map { |tag_taggable| tag_taggable.tag })
+  end
+
+  it 'on destroy destroys tag taggables, not taggables' do
```
sdjm-tessella, line 13: I wonder whether it would be advisable to put an underscore in `tag taggables` in these `it` descriptions so that they are easier to digest as a single entity?

spec/support/taggable.rb
Chris-Friend: I knew I'd done this somewhere

db/migrate/20200428064825_create_tag_taggables.rb
Chris-Friend: Good spot. Yes it should not be polymorphic

Chris-Friend:
> A few things to think about, but looking good overall. I did notice you have this set to merge into `gpl-426-allow-pretagging-samples` but I did create a branch that's just called `gpl-426`. I feel that by having the base branch for an issue just be called by its number, then it makes it clear that is the base branch, and anything with a description in the branch name is work that will be merged into that base branch. What do you think?

Chris-Friend:
> A few things to think about, but looking good overall. I did notice you have this set to merge into `gpl-426-allow-pretagging-samples` but I did create a branch that's just called `gpl-426`. I feel that by having the base branch for an issue just be called by its number, then it makes it clear that is the base branch, and anything with a description in the branch name is work that will be merged into that base branch. What do you think?

I will remove the branch I created for the same purpose as your gpl-426, and re-target this PR at that branch, maybe not in that order

spec/factories/tags.rb
```diff
@@ -3,5 +3,15 @@
     sequence(:oligo) { |n| ['A','C','G','T','A','C','G','T',n].shuffle[0,8].join }
     sequence(:group_id) { |n| n }
     tag_set_id { TagSet.find_or_create_by(name: 'Test Tag Set', uuid: '123456').id }
```
stevieing, line 5: Probably better to use a factory here. i.e. 
`create(:tag_set)`
I think you can just do
`tag_set { create(:tag_set) }`
The name of the tag set and the uuid are set as unique within the tag_set factory.

Chris-Friend: I did comment on this once already, but probably in the wrong place on GitHub

I have updated to use create(:tag_set) but have had to define the relationship still using tag_set_id as otherwise some requests/v1 tests fail for a reason I cannot determine

PR #343, opened 2020-04-29
WIP: Started GraphQl pooling samples

### LINKED ISSUE #213: GPL-386 As a user I need to be able to create ONT libraries
**User story**
As a user I need to be able to create ONT libraries 

Harry/Steve

**Acceptance criteria**
- api accept plate barcode and tag set name
- creates tube with a barcode
- creates library and link to tube
- add tag to existing request, link request to library
- can get list of all libraries

harrietc52:
Closes #213


app/graphql/types/mutation_types.rb
```diff
@@ -5,5 +5,7 @@ module Types
   class MutationTypes < BaseObject
     # Plates
     field :create_plate_with_ont_samples, mutation: Mutations::CreatePlateWithOntSamplesMutation
+
+    field :poolSamples, mutation: Mutations::PoolSamplesMutation
```
sdjm-tessella, line 9: The field should be snake_case and is the name of the method you supply when you request the mutation

app/graphql/types/outputs/library_type.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Outputs
+    # The type for Library objects.
+    class LibraryType < CommonOutputObject
+      field :id, Int, null: false
+      field :tubeBarcode, String, null: true
+      field :plateBarcode, String, null: true
+      field :pool, String, null: true
+      field :name, String, null: true
+      field :wells, String, null: true
+      field :tagSet, Integer, null: true
```
harrietc52, line 13: tag set can be determinded from the wells

app/graphql/mutations/pool_samples_mutation.rb
```diff
@@ -0,0 +1,15 @@
...
+module Mutations
+  # Mutation to handle the creation of a library(s), by pooling samples
+  class PoolSamplesMutation < BaseMutation
+    argument :arguments, Types::Inputs::PoolSamplesInputType, required: true
+
+    field :library, [Types::Outputs::LibraryType], null: true
+    field :errors, [String], null: false
+
+    # def resolve(arguments:)
+    #   debugger
```
harrietc52, line 12: - create tube with barcode
- create library
- add tag to existing request
- link request with library

app/graphql/types/outputs/library_type.rb
sdjm-tessella, line 13: All these fields should be specified in snake_case

app/graphql/types/query_types.rb
```diff
@@ -34,5 +34,26 @@ def wells(plate_id: nil)
...
+          name: 'TRAC-1-1-1', wells: 'A1-H3', tag_set: 24 },
+        { id: 2, tube_barcode: 'TRAC-2-2', plate_barcode: 'TRAC-1-1', pool: 2,
+          name: 'TRAC-1-1-2', wells: 'A4-H6', tag_set: 24 },
+        { id: 3, tube_barcode: 'TRAC-2-3', plate_barcode: 'TRAC-1-1', pool: 3,
+          name: 'TRAC-1-1-3', wells: 'A7-H9', tag_set: 24 },
+        { id: 4, tube_barcode: 'TRAC-2-4', plate_barcode: 'TRAC-1-1', pool: 4,
+          name: 'TRAC-1-1-4', wells: 'A10-H12', tag_set: 24 },
+        { id: 5, tube_barcode: 'TRAC-2-5', plate_barcode: 'TRAC-1-2', pool: 1,
+          name: 'TRAC-1-2-1', wells: 'A1-H12', tag_set: 96 }
+      ]
```
sdjm-tessella, line 56: I don't think GraphQL will cope well with this array being returned.  It uses the type of the value here to know which output type to use to create the JSON response.  We don't have an output type for arrays.  It might just handle this without an output type specified, but I can't be sure that's the case.

app/graphql/types/inputs/pool_samples_input_type.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Types
+  module Inputs
+    # The input arguments required to pool samples.
+    class PoolSamplesInputType < BaseInputObject
+      argument :plate_id, Int, required: true
+      argument :tag_set, Int, required: true # 24 or 96
```
sdjm-tessella, line 8: If we make both of these be required, you won't be able to use this input type to do "update" migrations in the future unless the user always updates both the plate ID and the tag set.  Maybe that's OK?  I avoided specifying that input fields be required so that updates could update just one value or the other.

app/graphql/types/mutation_types.rb
sdjm-tessella: I think it should also be more descriptive of what the mutation does and the class name for the mutation should match as well.  Something like `:create_ont_library_by_pool_samples`

PR #344, opened 2020-04-29
Gpl 426 move tubes

### LINKED ISSUE #339: GPL-nnn Move existing tube entities to exist as containers
**User story**
As a developer, I want to transition existing Tube entities in the ActiveRecord database so that they adopt the Container concern so that we can use tubes across the pipelines and also add the flexibility to register samples in tubes and plates for ONT and PacBio.

Stephen Inglis
Stuart McHattie

**Acceptance criteria**
1. Tubes exist as entities which adopt the Material concern.
2. Accessing the material property on a tube transparently returns the same object that it did before the transition.
3. Accessing a tube from a linked material transparently returns the same tube as before.
4. Factories that generate relationships between tubes and their materials create the relationship correctly via the ContainerMaterial joining object.

**Additional context**
Tube entities are currently not marked as Containers in the database schema.  This means they won't be able to take advantage of the ability for tubes to contain multiple tagged samples at once.  The existing Tube model has a direct relationship with Material, but this relationship will need to pass through a ContainerMaterial join object and so the database must be transitioned to maintain the existing links between Tubes and their Materials.

A database migration should be able to support this transition where a new ContainerMaterial is generated for each existing Tube and that object will hold the Material type and ID that the tube used to.  The Container on this object will reference the tube.  A new tube property on Pacbio Requests will be used as an alias to their container in the new relationship.

Factories will need to be updated so that new data coming in about Pacbio Requests in Tubes generate the new relationships between those entities where before it was a direct relationship.

This is a breakout story from GPL-426 #336

sdjm-tessella:
Closes #339

* Tubes moved from being attached to a Material to going via ContainerMaterial.
* Existing TubeMaterial concern updated to reflect the new relationship.
* Database migrated to match this.
* JSON:API Resources v1 endpoint updated to be as close as possible to how it was before.
* Tests updated for new assumptions around v1 endpoint.
* Rake tasks updated to be able to handle the new relationship for Pacbio and Saphyr data.


spec/requests/v1/saphyr/libraries_spec.rb
```diff
@@ -228,7 +232,8 @@
   context '#destroy' do
     context 'on success' do
       let!(:library) { create(:saphyr_library) }
-      let!(:tube) { create(:tube, material: library)}
+      let!(:tube) { create(:tube)}
+      let!(:container_material) { create(:container_material, container: tube, material: library)}
```
Chris-Friend, line 236: You do this a few times: create the tube them assign a material to it be creating a new container material. Would it be worth adding a `tube_with_material` factory method that takes a transient material and creates a container material? The existing tube factory methods, e.g. `tube_with_saphyr_request` could then subclass this maybe, passing in the correct material?

db/migrate/20200428092948_update_tubes_to_be_containers.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateTubesToBeContainers < ActiveRecord::Migration[6.0]
```
stevieing, line 1: Is there any risk of doing it like this? Could you get Harry to look at this just to make sure that this is going to work. We also have other team members with experience of migrating in production.

app/resources/v1/pacbio/container_material_resource.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
```
stevieing, line 2: Holy crap!!! I didn't realise you would have to do this. There is a lot of complexity to make this work which almost negates the point of polymorphism. We definitely need to move PacBio over to GraphQL. Hopefully this is an interim measure. Again could you get Harry to look at this as well.

app/resources/v1/saphyr/container_material_resource.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
```
stevieing, line 2: Again Holy crap!!! I didn't realise you would have to do this. There is a lot of complexity to make this work which almost negates the point of polymorphism. We definitely need to move Saphyr over to GraphQL. Hopefully this is an interim measure. Again could you get Harry to look at this as well.

app/models/concerns/tube_material.rb
```diff
@@ -3,8 +3,13 @@
 # Material
 module TubeMaterial
```
Chris-Friend, line 4: I don't know if you need this any more. Can you instead add something like: `has_one :tube, through: :container_material, source: :container, source_type: 'Tube'` to the material concern that will do the same thing? Or at least add this method to the material concern?

db/migrate/20200428092948_update_tubes_to_be_containers.rb
sdjm-tessella: I reviewed all the ways you can migrate data and this was the lowest risk I could find without introducing additional dependencies.  I read that using SQL directly and avoiding the model classes is the way forward.  Model classes might change name and this can break old migrations.  By referring to the tables directly, we know that table will always exist at this point in the migration history even if the model classes get renamed.

app/resources/v1/saphyr/container_material_resource.rb
sdjm-tessella: Yeah I wasn't proud of this, but I don't see another way while keeping the API the same.  I assume your comment is to say you like JSON:API even less now, rather than to ask me to change anything here?

app/models/concerns/tube_material.rb
sdjm-tessella: As discussed, I think this is a reasonable thing to leave in since it distinguishes materials that expect to only have a tube from those that are aware they have a polymorphic type of container.

app/resources/v1/saphyr/container_material_resource.rb
stevieing: I think me and JSON:API are just about to split up. I don't see the point in putting any more effort into something that is out of the door in a few weeks.

spec/models/pacbio/well_library_spec.rb
```diff
@@ -12,11 +12,10 @@
 
   it 'will have some library attributes' do
     request = create(:pacbio_request)
```
harrietc52, line 14: can remove this

app/models/pacbio/library_factory.rb
```diff
@@ -57,7 +58,9 @@ def build_library(library_attributes)
 
     def create_library(library_attributes)
       library_attributes_without_requests = library_attributes.except(:requests)
-      Pacbio::Library.new(library_attributes_without_requests.merge!(tube: Tube.new))
+      library = Pacbio::Library.new(library_attributes_without_requests)
+      @container_material = ContainerMaterial.new(container: Tube.new, material: library)
```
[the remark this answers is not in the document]
sdjm-tessella: I think generally it's an anti-pattern to make things public just for the sakes of testing.  Unit tests should only test the public interface of a class, not the inner workings. 

This test is supposed to check that the state of the factory is correct if we haven't yet called save, which it is doing.  It would be nice to check that the ContainerMaterial was created, but it doesn't matter that it has been until you save.  We check the link exists in the context checking the state after the save.

I've not wanted to change the existing code base, but another anti-pattern is getting the initializer to do too much work.  Initialize would, in an ideal world, only save values for the class to use.  There should be a method called `build` which does the building of the factory's output.  Then you separate object creation from the object's purpose.

PR #345, opened 2020-04-30
Gpl 426 many materials

Chris-Friend:

* Allow a container to have many materials


app/models/ont/request_factory.rb
```diff
@@ -0,0 +1,65 @@
...
+    validate :check_request
+
+    def initialize(attributes = {})
+      @well = attributes[:well]
+      return unless attributes.key?(:request_attributes)
+
+      build_request(attributes[:request_attributes])
+    end
+
+    attr_reader :well, :request, :well_request_join
```
sdjm-tessella, line 20: Do we want `:well` or `:well_request_join` to be accessed publicly?

app/models/ont/well_factory.rb
```diff
@@ -17,44 +17,29 @@ def initialize(attributes = {})
       build_well(attributes[:well_attributes])
     end
 
-    attr_reader :plate, :well, :request, :well_request_join
+    attr_reader :plate, :well
```
sdjm-tessella, line 20: Same question for `:plate` here.  Maybe it should be kept private.

app/models/ont/well_factory.rb
```diff
@@ -17,44 +17,29 @@ def initialize(attributes = {})
...
+
+    def request_factories
```
sdjm-tessella, line 22: And this feels like it should be private too.

spec/factories/plates.rb
```diff
@@ -1,4 +1,27 @@
...
+      end
+  
+      after :create do |plate, options|
+        options.well_count.times do |i|
+          create(:well, position: "A#{i}", plate: plate)
+        end
+      end
+    end
+
+    factory :plate_with_ont_samples do
```
sdjm-tessella, line 15: Good job getting these in here.  When I created the separate file I didn't know this was how it was supposed to be arranged.

spec/requests/v2/plates_spec.rb
```diff
@@ -121,7 +125,7 @@ def valid_query()
       expect(mutation_json['errors']).to contain_exactly('Wells {:message=>"This is a test error"}')
     end
 
-    def missing_required_fields_query()
+    def missing_required_fields_query
```
sdjm-tessella, line 128: Surprised that rubocop let me get away with that before

sdjm-tessella:
I suspect this needs some careful integration with the tubes being where they are now

PR #348, opened 2020-05-04
Gpl 386 ont library

Chris-Friend:

* Add Ont::Library models and database migration
* Add Ont library factory to generate libraries from a plate and tag set


spec/models/plate_spec.rb
```diff
@@ -5,4 +5,24 @@
     let(:labware_model) { :plate }
     it_behaves_like 'labware'
   end
+
+  it 'returns wells by column then row' do
+    plate = create(:plate_with_wells, row_count: 3, column_count: 12)
+    expected_positions = [
+      'A1', 'B1', 'C1', 'A2', 'B2', 'C2', 'A3', 'B3', 'C3', 'A4', 'B4', 'C4',
+      'A5', 'B5', 'C5', 'A6', 'B6', 'C6', 'A7', 'B7', 'C7', 'A8', 'B8', 'C8',
+      'A9', 'B9', 'C9', 'A10', 'B10', 'C10', 'A11', 'B11', 'C11', 'A12', 'B12', 'C12',]
```
sdjm-tessella, line 14: You have a stray final comma in this array

app/models/ont/library.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Ont
+  # Ont::Library
+  class Library < ApplicationRecord
+    include Material
+
+    has_many :requests, foreign_key: :ont_library_id, inverse_of: :library, dependent: :destroy
```
sdjm-tessella, line 8: Does this dependent relationship mean that if you destroy the library you also destroy the request?  Is that what we want?

Chris-Friend: I think this is what it does. I included this as on creating the library we modify the request, adding a tag to it. If we remove the library we still have a tagged request. Would a better option here be to include a join table from library to requests so you only delete the join table entry?

spec/factories/ont/libraries.rb
```diff
@@ -0,0 +1,18 @@
+FactoryBot.define do
+  factory :ont_library, class: 'Ont::Library' do
+    plate_barcode { "PLATE-1-123456" }
+    pool { 2 }
+    well_range { "A1-H8" }
```
sdjm-tessella, line 5: Aren't the wells `A1-H12` on these plates?

app/models/ont/library_factory.rb
```diff
@@ -0,0 +1,152 @@
...
+
+      validation_errors << ['Tag set', 'tag count must be a factor of plate well count']
+      false
+    end
+
+    def try_get_sorted_wells(primary_grouping_direction)
+      if primary_grouping_direction == 'column'
+        @sorted_wells = @plate.wells_by_column_then_row
+        true
+      elsif primary_grouping_direction == 'row'
```
sdjm-tessella, line 88: I was wondering whether these grouping directions should be `horizontal` and `vertical`?  I'm just thinking `row` and `column` infers that the grouping is per-row or per-column and it's not.

spec/factories/ont/libraries.rb
Chris-Friend: As this is a test factory, I'm not going to change this. I think the data added here is sufficient for testing purposes

app/models/ont/library_factory.rb
Chris-Friend: I'm happy to change these. I only think it makes sense to do so in the context of the factory though, i.e. the well still has a row and a column, and the plate still sorts by row then column etc.

sdjm-tessella: Yes absolutely.  The sorting of wells is row and column specific.  The creation of libraries is the right place for this.

app/models/ont/library.rb
sdjm-tessella: Is the tag singular and specific to the library?  If so, maybe it should exist on the library and you can just set this to null out the relationship and not modify the requests.  The request can delegate to the library tag if needed.

Chris-Friend: No, each well used to form a library has its requests tagged on library creation. All requests in a well will have the same additional tag

sdjm-tessella: OK sorry not sure if you've confirmed what I thought or not.  A library adds the same tag to each sample in the library?  So you could store that tag on the library instead of the request?

Chris-Friend: No. A library's requests come from many wells. The requests in each well used to form a library are tagged with a different tag

sdjm-tessella: Ah yes OK so although all the samples in a well have the same library tag, the tag used is different for each well in the library.  OK.  Well maybe the library tag on a request should be stored in a different field on the request so that you can separate it from the tags it has that weren't imposed by the library.

Chris-Friend: This is what we were discussing a few days ago when modelling out the relationships between Ont::Request and Ont::Library on Lucidcharts. Having multiple columns for tags was not desired

Chris-Friend: I admit updating the Ont::Request with new tags is also not a great solution as it ties the request to the library. This is handled for Pacbio by a three-way join between library, request and tags. That may be a better approach here too

sdjm-tessella:
> I know we were talking about the depency destroy relationship with requets, at the moment, ONT libraries cannot be deleted from the UI. There is a story in for that, but it is at the lower towards the bottom of the priority - not sure if that means we shouldnt add the correct support now though.. just a thought

I would imagine not setting up the join table now will make it trickier to do so in the future.

app/models/ont/library.rb
Chris-Friend: I have made the change to move to using a join table in my [next branch](https://github.com/sanger/traction-service/tree/GPL-386-ont-library-graphql)

Chris-Friend:
I have made the discussed changes, specifically updating the library information returned by graphql, and moving to use a join table, in my [next branch](https://github.com/sanger/traction-service/tree/GPL-386-ont-library-graphql)

PR #350, opened 2020-05-04
ONT dummy data

sdjm-tessella:
Adds plates for pooling scenarios 1 through 3 as dummy data.

Requests go through GraphQL.


lib/tasks/ont_runs.rake
```diff
@@ -0,0 +1,100 @@
...
+                 '   Use the RAILS_ROOT_URI environment variable to specify a different URI']
+  end
+
+  task create: :environment do
+    puts '-> Creating ONT runs using GraphQL'
+    puts '   Note these could take a few minutes to complete'
+    variables = OntPlates::Variables.new
+    submit_create_plate_query(description: 'pooling scenario 1', barcode: 'PLATE-PS01', wells: variables.wells(samples_per_well: 1))
+    submit_create_plate_query(description: 'pooling scenario 2', barcode: 'PLATE-PS02', wells: variables.wells(samples_per_well: 1))
+    submit_create_plate_query(description: 'pooling scenario 3', barcode: 'PLATE-PS03', wells: variables.wells(samples_per_well: 96))
```
Chris-Friend, line 88: I'd argue that these are 'plate input scenarios'. And I thought the 3 scenarios would be a plate with 1 sample in a well; 96 samples in a well; and 384 samples in a well?

lib/tasks/ont_runs.rake
```diff
@@ -0,0 +1,100 @@
...
+    puts '-> Successfully created all ONT runs'
+  end
+
+  task destroy: :environment do
+    # TODO: This needs to be made to work correctly for ONT
```
Chris-Friend, line 93: By this do you mean we need to review the model relationships in rails to ensure the models can be deleted in a sensible way, that this need to be implemented fully, or both?

lib/tasks/ont_runs.rake
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    puts "-> Succesfully created a plate for #{description}"
+  rescue Errno::ECONNREFUSED
+    show_errors ["-> Failed to connect to the Rails server at #{TractionGraphQL::RAILS_ROOT_URI}",
+                 '   Use the RAILS_ROOT_URI environment variable to specify a different URI']
+  end
+
+  task create: :environment do
+    puts '-> Creating ONT runs using GraphQL'
```
Chris-Friend, line 83: Will this rake task be used to create runs when implemented, as well as plates, or does it need re-describing?

lib/tasks/create_tags.rake
```diff
@@ -20,7 +52,21 @@ namespace :tags do
...
     Tag.create!(oligo: 'CTATACATAGTGATGTT', group_id: 'bc1021_BAK8B_OA', tag_set_id: set.id)
     Tag.create!(oligo: 'CACTCACGTGTGATATT', group_id: 'bc1022_BAK8B_OA', tag_set_id: set.id)
-    puts '-> Tags successfully created'
+    puts '-> Sequel_16_barcodes_v3 tags successfully created'
+
+    puts
+    puts '-> Creating ONT tag set for 96 sample wells'
+    tag_set = TagSet.create!(name: 'OntWell96Samples', uuid: SecureRandom.uuid)
+    puts '-> Tag Set successfully created'
+    oligo = +'ACGTACGTACGTACGT'
```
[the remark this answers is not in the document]
sdjm-tessella: `+` in front of a string tells Ruby you're OK with this string literal not being frozen, i.e. I want to mutate it and I know what I'm doing.  I think the + effectively makes the string be mutated upon creation so it's no longer really a literal.

lib/tasks/ont_runs.rake
sdjm-tessella: Mostly the latter.  If you look at the destroy method in other rake files, you'll see that they destroy specifically plates for their respective workflow.  I think this should be done at the end, if at all.

lib/tasks/ont_runs.rake
sdjm-tessella: When the endpoints are there to do more (like create libraries) then we can make this do what it says it does.

lib/tasks/ont_runs.rake
sdjm-tessella: Not sure what you mean.  The pooling scenarios require a plate of a specific set up.  Steve said not to worry about the 384 samples in a well.  The first two scenarios use a plate with 96 wells and a single sample in each.  The third scenario requires a plate with 96 wells, each containing either 96 or 384 samples.  We're just sticking to 96 for now.

Chris-Friend: What I'm getting at is that at the plate creation stage there is no distinction between scenarios 1 and 2. They both create a plate with one sample in a well. But I suppose you're setting up separate plates for the different scenarios, so require setting up plates for those scenarios

PR #356, opened 2020-05-06
Gpl 386 ont library graphql

Chris-Friend:
Please first review/merge https://github.com/sanger/traction-service/pull/348

* Fixes mentioned in https://github.com/sanger/traction-service/pull/348
** changing Library output fields through graphql
** adding a join table between library, request and tags
* Adding API endpoints for creating a new library and listing existing libraries


app/graphql/mutations/create_ont_libraries_mutation.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Mutations
+  # Mutation to handle the update of a Well.
+  class CreateOntLibrariesMutation < BaseMutation
+    argument :arguments, Types::Inputs::Ont::LibrariesInputType, required: true
+
+    field :tubes, [Types::Outputs::TubeType], null: false
```
sdjm-tessella, line 8: Not sure but I think the pattern for errors is that the object comes back `nil` rather than an empty array.

app/graphql/types/inputs/ont/libraries_input_type.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Types
+  module Inputs
+    module Ont
+      # The input arguments for Ont libraries.
+      class LibrariesInputType < BaseInputObject
```
sdjm-tessella, line 7: I think this name is a bit misleading since it infers you're providing information about a library, when you're actually specifying a plate and grouping strategy.  Maybe this should be called something like `LibraryCreationArgumentsInputType`.

app/graphql/types/query_types.rb
```diff
@@ -34,5 +34,15 @@ def wells(plate_id: nil)
...
+
+    # Ont
+
+    field :ont_libraries, [Types::Outputs::Ont::LibraryType], null: false do
+      description 'Find all Ont Libraries.'
+    end
+
+    def ont_libraries
+      Ont::Library.all
+    end
```
sdjm-tessella, line 46: I think the comment for the class on line 4 might need updating.  This must've been forgotten by me in a previous refactor

app/graphql/types/query_types.rb
```diff
@@ -34,5 +34,15 @@ def wells(plate_id: nil)
     def plates
       Plate.all
     end
+
+    # Ont
```
sdjm-tessella, line 38: Can you put more comments like this in further up?  i.e. `# Wells` above the well queries and `# Plates` above those.  Then change this one to `# Ont::Libraries`

app/models/ont/library_factory.rb
```diff
@@ -0,0 +1,172 @@
...
+
+      num_libraries.times do |lib_idx|
+        build_library(lib_idx, @tag_set.tags.count)
+      end
+    end
+
+    attr_reader :tubes
+
+    def save
+      return false unless valid?
```
sdjm-tessella, line 34: I think you need to add the `**option` argument to this `save` method and introduce the logic on this line that I used in my other branch.

app/models/ont/library_factory.rb
```diff
@@ -0,0 +1,172 @@
...
+module Ont
+  # LibraryFactory
+  # The factory will create libraries in tubes from a given plate
+  class LibraryFactory
+    include ActiveModel::Model
+
+    validate :check_validation_errors,
+             :check_libraries,
+             :check_library_requests,
+             :check_container_materials
```
sdjm-tessella, line 15: I think you may be missing a validation for `@tubes` here.

app/models/ont/library_request.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Ont
+  # Ont::LibraryRequest
+  class LibraryRequest < ApplicationRecord
+    belongs_to :library, foreign_key: :ont_library_id,
+                         inverse_of: :library_requests, dependent: :destroy
+    belongs_to :request, foreign_key: :ont_request_id, inverse_of: :library_requests
+    belongs_to :tag, foreign_key: :tag_id, inverse_of: false
```
sdjm-tessella, line 9: What does `inverse_of: false` do?

app/models/ont/library_factory.rb
Chris-Friend: I am not :). If I add a check on tubes they get validated twice, hence I do not explicitly validate

app/models/ont/library_request.rb
Chris-Friend: I'm guessing It specifies that the model to which this belongs does not need to define an inverse. I took this from some code elsewhere in the app. Will see if I can remove it

app/models/ont/library_factory.rb
sdjm-tessella: Maybe pop a comment in to help future developers (aka you) when they think the same thing I did

app/graphql/mutations/create_ont_libraries_mutation.rb
Chris-Friend: Change made to instead return nil

app/graphql/types/inputs/ont/libraries_input_type.rb
Chris-Friend: Input type renamed

app/graphql/types/query_types.rb
Chris-Friend: Done

app/models/ont/library_request.rb
Chris-Friend: Rubocop complains if an inverse_of isn't included 😞 _app/models/ont/library_request.rb:9:5: C: Rails/InverseOf: Specify an :inverse_of option._

app/models/concerns/labware.rb
```diff
@@ -18,6 +18,7 @@ def prefix
   def generate_barcode
     return if barcode.present?
 
-    update(barcode: "TRAC-#{self.class.prefix}-#{id}")
+    self.barcode = "TRAC-#{self.class.prefix}-#{id}"
+    save(validate: false)
```
sdjm-tessella, line 22: This probably deserves a comment as it is unusual

Chris-Friend: I was thinking about this and think I will try to instead stub out the Tube in the tests so this validation can still be performed

Chris-Friend: Addressed this by injecting my own object on creation so that any validation performed during object creation can be distinguished from validation performed in the factory

app/models/ont/library.rb
```diff
@@ -5,8 +5,29 @@ module Ont
...
+      return nil if plate_barcode.nil? || pool.nil?
+
+      "#{plate_barcode}-#{pool}"
+    end
+
+    def plate_barcode
+      name.delete_suffix("-#{pool}")
+    end
+
+    def tag_set_name
```
stevieing, line 23: You can do
`library_requests.first.&tag.&tag_set_name`

Chris-Friend:
> Looking good. I think the library factory looks like it could go through a bit of a refactor like with the request factory. Because nothing is user led we might not need most of the validation.

Would you like this refactoring as part of this PR? I'm not entirely sure what to refactor. Most of the validation is on the input parameters, that corresponding database entities do exist, and that they are consistent. Then all created entities are only validated once to reduce validation overhead.

One possibility would be to refactor out an `Ont::LibraryRequest` factory that accepts wells and a newly created library, but extracting a five line method to a new class seems like overkill compared to the benefits it will provide

app/models/ont/library.rb
```diff
@@ -5,8 +5,29 @@ module Ont
   class Library < ApplicationRecord
     include Material
 
-    has_many :requests, foreign_key: :ont_library_id, inverse_of: :library, dependent: :destroy
+    has_many :library_requests, foreign_key: :ont_library_id,
+                                inverse_of: :library, dependent: :destroy
 
-    validates :name, :pool, :well_range, :pool_size, presence: true
+    validates :name, :pool, :pool_size, presence: true
```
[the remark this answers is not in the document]
Chris-Friend: A pool is generated when creating a library. I'd have thought a library isn't valid unless you specify it's pool number on a plate, even if that plate only goes to make a single pool?

app/graphql/types/outputs/ont/library_type.rb
```diff
@@ -0,0 +1,17 @@
...
+    module Ont
+      # The type for Ont::Library objects.
+      class LibraryType < CommonOutputObject
+        field :name, String, null: false
+        field :plate_barcode, String, null: false
+        field :pool, Integer, null: false
+        field :pool_size, Integer, null: false
+        field :tag_set_name, String, null: true
+        field :tube_barcode, String, null: true
+      end
```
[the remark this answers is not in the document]
Chris-Friend: As discussed `well_primary_grouping_direction` removed from the input type, but `pool_size` kept as it will remain available on the active model, and will probably vary with further implementations of Ont workflows

PR #358, opened 2020-05-07
Gpl 387 active records

Chris-Friend:

* Add Ont::Run active model
* Add Ont::Flowcell active model

Main points to look at:
* are the relationships between library - flowcell - run. Do the dependencies make sense? E.g. destroying a run on destroying a flowcell
* have I included all the expected fields?
* should an Ont::Run require an instrument_name. It is currently optional


db/migrate/20200507071910_create_ont_flowcells.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOntFlowcells < ActiveRecord::Migration[6.0]
+  def change
+    create_table :ont_flowcells do |t|
+      t.integer :position
+      t.string :uuid
+      t.belongs_to :ont_run, index: true
+      t.belongs_to :ont_library, index: true
```
sdjm-tessella, line 7: Is this the right way around for this relationship?  I would have thought the flow cell owned the ONT library?

Chris-Friend: I'm glad you picked up on this 😄. I did have it the way round you suggest originally, but implementing it such that a flowcell belongs to a library means you don't have to specify anything extra for the library -> flowcell relationship to be optional, but the flowcell -> library relationship to be required. I don't think there is a great amount of extra gained from having it the other way round, apart from it actually making sense on reading. Perhaps I should add a comment?

sdjm-tessella: I suppose it doesn't make a lot of difference.  You either have the ID in one table or the other table.  In any case, the relationship exists.  Probably leave it as you have it.

PR #360, opened 2020-05-07
Gpl 387 run factory

Chris-Friend:

* Add the run factory
* Ensure it only validates it's created entities once, and validation can be skipped altogether

### DIFF
--- a/app/models/ont/run_factory.rb
+++ b/app/models/ont/run_factory.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+# Ont namespace
+module Ont
+  # RunFactory
+  # The factory will create runs from a list of libraries
+  class RunFactory
+    include ActiveModel::Model
+
+    validate :check_run, :check_flowcells
+
+    def initialize(library_names = [])
+      @flowcells = []
+      build_run(library_names)
+    end
+
+    attr_reader :run
+
+    def save(**options)
+      return false unless options[:validate] == false || valid?
+
+      @run.save(validate: false)
+      @flowcells.each { |flowcell| flowcell.save(validate: false) }
+      true
+    end
+
+    private
+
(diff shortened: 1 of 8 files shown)


app/models/ont/run_factory.rb
```diff
@@ -0,0 +1,62 @@
...
+  # RunFactory
+  # The factory will create runs from a list of libraries
+  class RunFactory
+    include ActiveModel::Model
+
+    validate :check_run, :check_flowcells
+
+    def initialize(library_names = [])
+      @flowcells = []
+      build_run(library_names)
```
harrietc52, line 14: I think the RunFactory is going to be initialized with libraryNames and position, a FlowcellType input, which includes the position. So `[{position: 1, library_name: 'aName'}, {position: 2, library_name: 'anotherName'}]`

app/models/ont/run_factory.rb
```diff
@@ -0,0 +1,62 @@
...
+      @run.save(validate: false)
+      @flowcells.each { |flowcell| flowcell.save(validate: false) }
+      true
+    end
+
+    private
+
+    attr_reader :flowcells
+
+    def build_run(library_names)
```
harrietc52, line 31: Which would change the was `idx` is used here as the position, as it is being passed in

config/pipelines.yml
```diff
@@ -184,6 +184,7 @@ default: &default
     covid:
       external_study_id: "example id"
       species: "example species"
+      instrument_name: "GridIon"
```
harrietc52, line 187: "GridION"

app/models/ont/run_factory.rb
stevieing: Agreed. I thought that we would have a flowcell type with a position and a library name attribute.

PR #361, opened 2020-05-11
Gpl 387 graphql endpoints

sdjm-tessella:
Needs #360 to be merged before this one can be reviewed separately.

* Adds an endpoint for creating ONT runs from libraries.
* Adds output types to GraphQL to show the created objects from the above endpoint.
* Adds tests for the creation process, mocking out the factory.


app/graphql/mutations/create_covid_run_mutation.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Mutations
+  # Mutation to create a COVID run.
+  class CreateCovidRunMutation < BaseMutation
+    argument :flowcells, [Types::Inputs::Ont::FlowcellInputType], required: true
+
+    field :run, Types::Outputs::Ont::RunType, null: true
+    field :errors, [String], null: false
+
+    def resolve(flowcells:)
```
harrietc52, line 11: What does `flowcells:` do - with the colon?

app/graphql/types/outputs/ont/flowcell_type.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Types
+  module Outputs
+    module Ont
+      # The type for Ont::Flowcell objects.
+      class FlowcellType < CommonOutputObject
+        field :position, Integer, null: false
+        field :uuid, String, null: false
```
harrietc52, line 9: I don't think the UI requires the uuid, but cant harm in keeping it in the api?

app/models/ont/run_factory.rb
```diff
@@ -0,0 +1,62 @@
...
+  # RunFactory
+  # The factory will create runs from a list of libraries
+  class RunFactory
+    include ActiveModel::Model
+
+    validate :check_run, :check_flowcells
+
+    def initialize(library_names = [])
+      @flowcells = []
+      build_run(library_names)
```
harrietc52, line 14: See position comment from #360

spec/models/ont/run_factory_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+      allow_any_instance_of(Ont::Flowcell).to receive(:errors).and_return(errors)
+      factory = Ont::RunFactory.new(['PLATE-2-1234-2'])
+      expect(factory).to_not be_valid
+    end
+  end
+
+  context '#save' do
+    context 'valid build' do
+      context 'with no library names' do
+        it 'creates a run' do
```
harrietc52, line 33: Should it still create a run if no library names are provided. I would have thought the flowcell validation would fail as it doesnt have a library?

app/models/ont/run_factory.rb
```diff
@@ -0,0 +1,62 @@
...
+module Ont
+  # RunFactory
+  # The factory will create runs from a list of libraries
+  class RunFactory
+    include ActiveModel::Model
+
+    validate :check_run, :check_flowcells
+
+    def initialize(library_names = [])
+      @flowcells = []
```
stevieing, line 13: you can turn this into an attr_reader
```
def flowcells
  @flowcells ||= []
end
```

app/models/ont/run_factory.rb
```diff
@@ -0,0 +1,62 @@
...
+      @run.save(validate: false)
+      @flowcells.each { |flowcell| flowcell.save(validate: false) }
+      true
+    end
+
+    private
+
+    attr_reader :flowcells
+
+    def build_run(library_names)
```
stevieing, line 31: Should this accept the position rather than generating it from an index?

app/graphql/mutations/create_covid_run_mutation.rb
sdjm-tessella: When you define a method and you put a colon after an argument name it becomes a named argument.  When this method is called by the GraphQL library, it expects to be able to pass the fields as named arguments with the same name as the field.

app/graphql/types/outputs/ont/flowcell_type.rb
sdjm-tessella: Sure .. I was just putting the fields from the underlying Rails model so you can get them if you want and, if you don't, you don't ask for them.

app/models/ont/run_factory.rb
sdjm-tessella: This was because the destination branch hadn't been merged back into this one yet.  I've done this and that should resolve your concerns.

app/models/ont/run_factory.rb
sdjm-tessella: This was because the destination branch hadn't been merged back into this one yet.  I've done this and that should resolve your concerns.

app/models/ont/run_factory.rb
sdjm-tessella: I've updated this

spec/models/ont/run_factory_spec.rb
sdjm-tessella: I don't know .. what does the UI send?  I can see why it doesn't harm to do this as well as it being a problem.  If the UI sends the run to the backend before the user has chosen libraries then updates the run as they drag in each library, then we probably do want the behaviour here.  If it only sends the run once the user has added some libraries then we could make the run factory reject the new run.  I didn't write this code, it just hadn't been merged back from the destination branch before you spotted this.  We should ask @Chris-Friend why he did it this way.

Chris-Friend: I did this as I assumed a run was valid without any flowcells. It doesn't necessarily make sense in the domain, but in terms of code implementation and user workflows, as Stuart suggests, it gives us more flexibility on how runs are created

app/models/ont/run.rb
```diff
@@ -7,6 +7,8 @@ class Run < ApplicationRecord
 
     has_many :flowcells, foreign_key: :ont_run_id, inverse_of: :run, dependent: :destroy
 
+    validates :instrument_name, presence: true
```
stevieing, line 10: Sorry. I should have spotted this earlier but again instrument name will be hardcoded.

Issue #362, opened 2020-05-11
GPL-457 As developers we need to modify the creation of samples to reflect the simplified scenario for creating ONT samples


stevieing:
**User story**
As developers we need to modify the creation of samples to reflect the simplified scenario for creating ONT samples

**Who the primary contacts are for this story**
Steve / Harry

**Acceptance criteria**
To be considered successful the solution must allow:
1. A plate will contain 96 wells
2. Each well will contain a single sample
3. Each sample must be pre-tagged with a pre existing PCR tag
4. A sample will only be sequenced once

**Additional context**

Chris-Friend:
In order to support this in the most straightforward way possible, we will make the following changes:
* remove sample request relationship
* In the request factory validate that samples have a valid tag, and assign to them this tag
* In the well factory ensure that each well only has a single sample (retaining flexibility to allow multiple samples in a single well)

Tag validation (that all tags are unique) will now be done on library creation.

PR #364, opened 2020-05-11
Remove ont sample relationship

Chris-Friend:
Part of #362

* Remove the relationship to sample from Ont::Request
* Have the Ont::Request store the sample name and external_id, but not the external_study_id
* Add a database migration to this effect
* Update the GraphQL endpoint, note that you mutate with a sample, but get back a request
* Update the request factory to require each sample has a tag
* Update the well factory to only allow a single sample in a given well, and not check the uniqueness of a samples

Note I have kept the TagService as this may be of use when updating the library factory. If not I will remove it then


app/models/ont/well_factory.rb
```diff
@@ -37,28 +37,17 @@ def build_well(attributes)
...
-        ::TagService.new(nil)
-      when 96
-        ::TagService.new(::TagSet.find_by!(name: 'OntWell96Samples'))
-      when 384
-        ::TagService.new(::TagSet.find_by!(name: 'OntWell384Samples'))
-      else
-        raise "'#{num_samples}' is not a supported number of samples"
-      end
+    def validate_num_samples(num_samples)
+      raise "'#{num_samples}' is not a supported number of samples" unless num_samples == 1
```
sdjm-tessella, line 50: I don't think we need to be raising exceptions here.  You can add the validation for the number of samples in `check_request_factories` to ensure there's only one and add a error message there instead of throwing an exception and having to catch it again.  You can probably do away with the `@raised_exceptions` workflow then.

Or, even simpler, don't accept an array of samples.  We can make GraphQL only accept one sample.

app/models/ont/request_factory.rb
```diff
@@ -8,73 +8,59 @@ module Ont
...
-    attr_reader :request
+    attr_reader :ont_request
 
     def save(**options)
       return false unless options[:validate] == false || valid?
 
       # No need to validate any lower level objects since validation above has already checked them
-      request.save(validate: false)
-      tag_taggable&.save(validate: false)
+      ont_request.save(validate: false)
```
stevieing, line 26: Can we now wrap this in a transaction as each object is singular?

app/models/ont/well_factory.rb
```diff
@@ -8,7 +8,7 @@ module Ont
   class WellFactory
     include ActiveModel::Model
 
-    validate :check_well, :check_request_factories, :check_tag_service, :check_for_raised_exceptions
+    validate :check_well, :check_request_factories, :check_for_raised_exceptions
 
     def initialize(attributes = {})
       @request_factories = []
```
stevieing, line 14: we can change this so that @request_factories becomes an attr_reader i.e. 
```
def request_factories
@request_factories ||= []
end
```

app/models/ont/well_factory.rb
```diff
@@ -37,28 +37,17 @@ def build_well(attributes)
       return unless attributes.key?(:samples)
 
       begin
-        @tag_service = create_tag_service(attributes[:samples].count)
+        validate_num_samples(attributes[:samples].count)
         @request_factories = attributes[:samples].map do |request_attributes|
```
stevieing, line 41: Further to above you can change this is
```
attributes[:samples].map do |request_attributes|
  request_factories << RequestFactory.new(well: well, request_attributes: request_attributes)
end
```

app/models/ont/request_factory.rb
Chris-Friend: Do you mean something like:
```
ActiveRecord::Base.transaction do
  ont_request.save(validate: false)
  tag_taggable.save(validate: false)
  well_request_join.save(validate: false)
end
```

Chris-Friend: What benefit does this have? Should we be wrapping all factory saves in a transaction like this?

app/models/ont/well_factory.rb
Chris-Friend: Resolved my making the well factory only have at most a single request factory

app/models/ont/well_factory.rb
Chris-Friend: Resolved as above

app/models/ont/request_factory.rb
sdjm-tessella: It would not benefit in terms of performance, but it would (assuming it works) ensure that if any save fails, the data gets rolled back.  We aren't asking for validation on each save, so I suspect the save would only fail if the database fails to respond to our request.

stevieing: Yes. Exactly what I meant.

PR #368, opened 2020-05-12
Gpl 386 ont library changes

Chris-Friend:
Links to #213 

* Remove request libraries join model
* Check the uniqueness of tags in the library factory
* Simplify the library factory and graph ql input type


app/models/ont/library_factory.rb
```diff
@@ -1,183 +1,99 @@
...
 
     def save(**options)
       return false unless options[:validate] == false || valid?
 
-      @libraries.each { |lib| lib.save(validate: false) }
-      @library_requests.each { |lib_req| lib_req.save(validate: false) }
-      @tubes.each { |tube| tube.save(validate: false) }
-      @container_materials.each { |cont_mat| cont_mat.save(validate: false) }
+      @library.save(validate: false)
+      @tube.save(validate: false)
```
sdjm-tessella, line 24: We have a getter for all these instance variables so we should drop the `@` when only getting the variable.

sdjm-tessella: I know I've flip flopped on this decision in the past, but I read up GPGs on this and you should use the non-@ version and use getters and/or setters instead via `attr_reader`, `attr_writer` and `attr_accessor`, scoping those appropriately (like you already have)

spec/factories/ont/requests.rb
```diff
@@ -3,7 +3,7 @@
     name { 'request name' }
     external_id { 'request external id' }
 
-    factory :ont_request_with_tags do
```
sdjm-tessella, line 6: :+1:

sdjm-tessella:
:+1:

Issue #215, opened 2020-01-13
GPL-387 As a user I need to be able to create an ONT run


stevieing:
**User story**
As a user I need to be able to create an ONT run

**Who the primary contacts are for this story**
Harry/Steve

**Acceptance criteria**
To be considered successful the solution must allow:

Run required fields:
- Instrument name
- Experiment name
- State

A run has many flowcells. Flowcell required fields:
- Flowcell ID
- Flowcell position
- Kit selection
- Basecall model (High accuracy / Fast / No basecalling)

A flowcell has a library
Library required fields:
...

- Successfully created a pacbio run in the database
- can GET run/runs

**Additional context**
What events are created?
Any automation with ONT API?

## Breakdown of tasks

* Generate architecture of active record models.
* Create a factory for generating runs.
* Create GraphQL endpoints.
  * Retrieve libraries with filter to exclude those already associated with a flowcell.
  * Mutation to create runs.
* Update message specifications for ONT runs.
* Message over Rabbit when runs are created as part of the above mutation.

harrietc52:
On run creation
- set experiment name
- create flowcells
- each flowcell have a uuid, position
- each flowcell is associated with a library

Unknowns
- store instrument name?
- store state?

Chris-Friend:
[GPL-387](https://github.com/sanger/traction-service/tree/GPL-387) is the feature branch for this story. The storyy will be complete on merging this feature branch into develop

Tasks breakdown:
* _Add Ont::Run and Ont::Flowcell active models_. The run should store the instrument name (fixed and from config), the experiment name (id), state (see Saphyr run, e.g. 'pending'), and should have many flowcells. The flowcell a position, a uuid and must have a single library. The library relationship to a flowcell should be optional
* _Add run and flowcells factories_. This should take a list of library identifiers (names or ids) and generate a single run with many flowcells containing a single library
* _Add new GraphQL endpoints_. One to create an ont run. Update the existing libraries endpoint to allow returning only libraries that have not been assigned to a flowcell
* _Send flowcells message(s) to the MLWH_. This will require adding a new ont flowcells message config in the `pipelines.yml`, and should be called in the GraphQL controller after successful creation of the run. Generated message should be tested

stevieing:
The message to the warehouse should contain the following fields. For each flowcell there will be a separate entry for each request in a library:

id_flowcell_lims: flowcell.id,
updated_at: timestamp,
sample_uuid: sample.external_id,
[...]

PR #372, opened 2020-05-12
Gpl 387 ont message spec

Chris-Friend:
Links to #215 

* Add Ont Covid message spec, for use with a single Ont::Request

I know this isn't complete , but I thought the best way to get feedback on it would be to submit a PR


spec/pipelines/ont/message_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    # pipeline_id_lims:
+    # library_preparation_type:
+    # tag_identifier:
+    # tag_sequence:
+    # tag_set_id_lims:
+    # tag_set_name:
+    # tag2_identifier:
+    # tag2_sequence:
+    # tag2_set_id_lims:
+    # tag2_set_name:
```
sdjm-tessella, line 62: If these won't be added in this PR, I would suggest you remove them

Chris-Friend: I've removed these comments

config/pipelines.yml
```diff
@@ -180,6 +180,41 @@ default: &default
...
+          # pipeline_id_lims:
+          # library_preparation_type:
+          # tag_identifier:
+          # tag_sequence:
+          # tag_set_id_lims:
+          # tag_set_name:
+          # tag2_identifier:
+          # tag2_sequence:
+          # tag2_set_id_lims:
+          # tag2_set_name:
```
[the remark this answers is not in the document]
Chris-Friend: I've added a TODO to add a bit more context on why these comments are here

Chris-Friend:
The below is the message received for one of the requests:

{"lims":"Traction","oseq_flowcell":{"last_updated":"2020-05-13T13:40:28.271Z","id_flowcell_lims":18,"sample_uuid":"H12-ExtId","study_uuid":"test study id","experiment_name":7,"instrument_name":"GridION","instrument_slot":2,"pipeline_id_lims":"test value","tag_identifier":112,"tag_sequence":"ACGTACGTACGTCTGG","tag_set_id_lims":2,"tag_set_name":"OntWell96Samples","tag2_identifier":null,"tag2_sequence":null,"tag2_set_id_lims":null,"tag2_set_name":null}}

app/graphql/mutations/create_covid_run_mutation.rb
```diff
@@ -15,10 +15,19 @@ def resolve(flowcells:)
       factory = Ont::RunFactory.new(flowcells.to_a)
 
       if factory.save
+        send_messages(run: factory.run)
```
stevieing, line 18: Should this be the other way round? Not sure just asking the question.

app/messages/messages/message.rb
```diff
@@ -63,8 +64,11 @@ def instance_value(object, field)
                               field[:value].split('.')[1..-1])
       when :array
         build_children(object, field)
+      when :nil
```
stevieing, line 67: I don't know if we need this. See comments below.

config/pipelines.yml
```diff
@@ -180,10 +180,56 @@ default: &default
...
+          tag_sequence:
+            type: :model
+            value: tags.first.oligo
+          tag_set_id_lims: 
+            type: :model
+            value: tags.first.tag_set.id
+          tag_set_name:
+            type: :model
+            value: tags.first.tag_set.name
+          tag2_identifier: &nil
```
stevieing, line 224: I think you can just do but not sure:
`You can just do tags.&last.&tag_set.&name`
There might be another way where we could use the children method but that might not work because of the tag and tag2 thing.

app/graphql/mutations/create_covid_run_mutation.rb
Chris-Friend: What do you mean? Asking the run factory to send messages?

stevieing: As in putting send_messages after returning the response. I don't know if that is possible.

config/pipelines.yml
Chris-Friend: Would tags&.last not just return the first tag?

stevieing: Yes you're right it would. tags[1] wouldn't work either. I think you can just return an empty string rather than adding extra logic. I will double check.

app/graphql/mutations/create_covid_run_mutation.rb
Chris-Friend: I also do not think this is possible. Is it an issue if we send the message first? This seems to be how things are done in the saphyr and pacbio controllers

stevieing: We do that in the other pipelines too but now I have realised you can add an after hook! Not really a problem if it can't be changed just thought it might slow the render down. Not really a problem.

PR #376, opened 2020-05-12
Rubocop pre commit hook

sdjm-tessella:
A pre-commit hook that runs rubocop on your modified and added files before the commit is accepted.  See the README for how to enable this, or for those who don't want to enable it, you don't have to if you don't follow the instructions in the README.


.githooks/pre-commit
```diff
@@ -0,0 +1,26 @@
...
+
+changed_files = `git status --porcelain`.split(/\n/)
+unstaged_files = `git ls-files -m`.split(/\n/)
+
+changed_files = changed_files.map do |file|
+  /^\s*[AM]+\s+(?<filename>.+)$/ =~ file # Find added or removed files
+  filename
+end.compact
+
+changed_files -= (unstaged_files - changed_files)
```
Chris-Friend, line 13: Is this right? Does it not evaluate to `changed_files = 2 * changed_files - unstaged_files`?

sdjm-tessella: I did think this line looked dodgy.  I've modified the script to have better variable names as well.

PR #378, opened 2020-05-13
Gpl 387 create run rake

Chris-Friend:
Links to #215 

Please first review https://github.com/sanger/traction-service/pull/361 as this PR continues on from that

* Update ont_runs:create rake task to create runs
* Ensure that created runs and flowcells are deleted on calling ont_runs:destroy


lib/traction_graphql.rb
```diff
@@ -91,3 +91,29 @@ module OntLibraries
...
+  GRAPHQL
+
+  # Methods to create variable objects for GraphQL
+  class Variables
+    def flowcells(library_names:)
+      library_names.each_with_index.map do |library_name, idx|
+        flowcell library_name: library_name, position: idx + 1
+      end
+    end
+
```
sdjm-tessella, line 111: I think you can stick the word `private` in here

lib/tasks/ont_runs.rake
```diff
@@ -83,3 +83,37 @@ def create_libraries(count:)
 
   puts '-> Successfully created ONT libraries'
 end
+
+def submit_create_run_query(variables:, library_names:)
+  flowcells = variables.flowcells(library_names: library_names)
```
sdjm-tessella, line 88: Instead of passing the `variables` object in here then just calling one method on it, would it be tidier to call this method in `create_runs` and pass in the `flowcells` instead of `variables`?

Chris-Friend: Try telling that to rubucop 🙄. I'll give it another go

Chris-Friend: I'm choosing not to fix this as rubocop is rejecting any of my attempts

PR #380, opened 2020-05-13
GPL-387: Update GraphQL queries

sdjm-tessella:
Part of #215

* Add a filter spec for libraries endpoint.
* Add an ont_runs endpoint for getting all runs.
* Add an ont_run endpoint for getting a single run by ID

### DIFF
--- a/spec/requests/v2/ont/runs_spec.rb
+++ b/spec/requests/v2/ont/runs_spec.rb
@@ -1,27 +1,56 @@
 # frozen_string_literal: true
 
-require "rails_helper"
+require 'rails_helper'
 
 RSpec.describe 'GraphQL', type: :request do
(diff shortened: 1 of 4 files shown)


spec/requests/v2/ont/runs_spec.rb
```diff
@@ -1,27 +1,80 @@
 # frozen_string_literal: true
 
-require "rails_helper"
+require 'rails_helper'
 
 RSpec.describe 'GraphQL', type: :request do
-  let(:run_factory) { instance_double("Ont::RunFactory") }
+  context 'get run' do
+    context 'when there is a valid run' do
```
Chris-Friend, line 7: Does it make sense to also test 'when there is not a valid run' since you've split these into the opposite context?

PR #386, opened 2020-05-14
GPL-388: Edit existing runs

### LINKED ISSUE #317: GPL-388 As a user I need to be able to edit ONT runs
**User story**
As a user I need to be able to edit ONT runs 

Steve/ Harry

**Acceptance criteria**
1. Changing the state of an existing run.
2. Replacing the flowcells (associated libraries) of an existing run.

**Additional context**
This can be implemented as a single GraphQL mutation endpoint for updating a run.  The run will be identified by its `id` and can alter the `state` or the entire list of `flowcells`.  The existing flowcells will be destroyed and replaced with newly generated ones.

We also need to consider which event messages are sent over RabbitMQ.  Libraries/Samples that are moved to another flowcell will be easy for the warehouse to identify as being updated if we only send messages for their new location.  But Libraries/Samples that are removed from the run will not be so easy to identify if messages aren't created for them.

sdjm-tessella:
Closes #317

This has a followup pull request #387 which updates the GraphQL documentation for these changes.

* Add a mutation for editing existing runs.
* Update the RunFactory to support modifying an existing run, or creating one if it isn't supplied.
* Add tests for all of the above.


harrietc52:
I avent had a proper look though, I was just using it to start the UI edit runs. But please could there be a query type to get all ont_runs, and get one ont run by id, if there isnt already.

spec/models/ont/run_factory_spec.rb
```diff
@@ -133,7 +239,11 @@
 
           it 'does not validate any flowcells' do
             allow(Ont::Flowcell).to receive(:new).and_return(flowcell)
-            expect(flowcell).to_not receive(:valid?)
+            # TODO: Ideally we'd not validate any flow cells,
```
Chris-Friend, line 242: Why don't we need to validate (the new) flowcells explicitly? Not doing so allows you to submit a library_name that does not exist. Does validating the run validate the flowcells by association?

app/models/ont/run_factory.rb
```diff
@@ -19,46 +21,33 @@ def save(**options)
...
 
-      return if @run.valid?
+      return if run.valid?
 
       run.errors.each do |k, v|
         errors.add(k, v)
       end
     end
-
-    def check_flowcells
```
Chris-Friend, line 54: Again I don't know why this validation is removed, unless validating the run validated the flowcells. If so could it be commented somewhere? And a test added to the run_spec.rb?

spec/models/ont/run_factory_spec.rb
sdjm-tessella: Yes it does ... this is proven by the unit tests you added around validation.

Chris-Friend: Should checking the run validates flowcells be in the run_spec tests instead of this run factory?

I'm not following why this test should expect flowcells to be validated when we choose not to validate the run, and don't explicitly validate any flowcells in the run factory

app/graphql/mutations/update_covid_run_mutation.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    private
+
+    def update_run_state(run, state)
+      run.update(state: state) unless state.nil?
+    end
+
+    def update_run_flowcells(run, flowcell_specs)
+      return if flowcell_specs.nil? || flowcell_specs.count == 0
```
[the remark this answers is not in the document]
sdjm-tessella: I guess you'll be disappointed.  I'm happy to change this if we think that behaviour would be more desirable.  In reality, I don't think this will happen, otherwise you'll be left with runs that have no flowcells and are therefore invalid for submission. There would be nothing to submit a message for the run.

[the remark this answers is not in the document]
Chris-Friend: But can I not create a run with no flowcells using the `CreateCovidRunMutation`? It seems contradictory that you can create a run with no flowcells, but not update a run to have no flowcells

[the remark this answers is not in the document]
Chris-Friend: It may be that the creation workflow needs changing to not allow creating an empty run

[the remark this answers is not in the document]
sdjm-tessella: I'll make it generate an error if you pass in an empty array.  Only accepting a populated array or nil

app/graphql/mutations/update_covid_run_mutation.rb
```diff
@@ -0,0 +1,45 @@
...
+    argument :id, ID, 'The ID of the Ont Run to update.', required: true
+    argument :properties, Types::Inputs::Ont::RunInputType, 'The properties to update on the run.',
+             required: true
+
+    field :run, Types::Outputs::Ont::RunType, 'The updated Run, or null if errors were thrown.',
+          null: true
+    field :errors, [String], 'An array of error messages thrown while updating the run.',
+          null: false
+
+    def resolve(id:, properties:)
```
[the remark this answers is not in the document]
sdjm-tessella: :+1:

app/models/ont/run_factory.rb
sdjm-tessella: There are tests already that prove this happens.  I don't feel it's justified to add a comment because we don't keep an array in the factory of these objects so it doesn't even feel like they should be validated.  They're stored directly on the run instead so we should expect the run to validate them when we validate it.

Chris-Friend: Since the run validates the flowcells, do we need the run factory tests that check the validation on the flowcell? I guess we do since we wan't to ensure it isn't over-validated

Chris-Friend: Resolving this in favour of the above conversation about the same thing

PR #387, opened 2020-05-14
GPL-388: Update GraphQL documentation

### LINKED ISSUE #317: GPL-388 As a user I need to be able to edit ONT runs

sdjm-tessella:
Closes #317

This can wait for #386 to be merged before it can be merged on top.

* Updated GraphQL documentation after recent changes for editing a run


Chris-Friend:
Is there any documentation on how to update the GraphQL documentation? In the readme for example?

PR #389, opened 2020-05-18
Modify ont rake task

stevieing:
Closes #

* modify rake task for creating ont runs so that you can create plates and libraries to make it easier to test with the ui
* added a count so you can specify the number of plates and runs you would like to create
* modified so that you can rerun it without getting any errors. It will take the last plate and set the barcode based on that.
* rubocopped so split up the task a bit.

### DIFF
--- a/lib/tasks/ont_data.rake
+++ b/lib/tasks/ont_data.rake
@@ -75,7 +75,7 @@ def create_number_of_plates(count, plate_no)
   variables = OntPlates::Variables.new
   constants_accessor = Pipelines::ConstantsAccessor.new(Pipelines.ont.covid)
 
-  plate_no.upto(plate_no+count-1).collect do |i|
+  plate_no.upto(plate_no + count - 1).collect do |i|
     create_plate(i, variables, constants_accessor)
   end
 end
--- a/lib/tasks/ont_data.rake
+++ b/lib/tasks/ont_data.rake
@@ -3,14 +3,14 @@
 require_relative '../traction_graphql'
 
 namespace :ont_data do
-  task :create, [:num] => :environment do |t, args|
+  task :create, [:num] => :environment do |_t, args|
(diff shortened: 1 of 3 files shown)


lib/tasks/ont_data.rake
```diff
@@ -2,24 +2,43 @@
...
+    count = args[:num]&.to_i || 5
+
+    # create count plates
+    barcodes = create_plates(count: count)
+    Plate.where(barcode: barcodes)
+
+    # create count plates with libraries
+    barcodes = create_plates(count: count)
+    plates = Plate.where(barcode: barcodes)
+    create_libraries(plates: plates)
```
sdjm-tessella, line 18: It seems like you don't need to look the plates up for this?  It only needs the plate barcodes which you already had.  I don't really mind that you're passing the whole plate in here, but then in `create_runs` further down you pass in the library names and not the libraries themselves, so there's some inconsistency.

lib/tasks/ont_data.rake
```diff
@@ -46,18 +65,33 @@ end
...
   variables = OntPlates::Variables.new
   constants_accessor = Pipelines::ConstantsAccessor.new(Pipelines.ont.covid)
-
-  count.times do |i|
-    plate_no = i + 1
-    submit_create_plate_query(plate_no: plate_no,
-                              barcode: "DEMO-PLATE-#{plate_no}",
-                              wells: variables.wells(sample_name: "for Demo Plate #{plate_no}",
-                                                     constants_accessor: constants_accessor))
+  [].tap do |barcodes|
```
sdjm-tessella, line 80: I suspect this block could be written more clearly as:

```
barcodes = (0..count).map do |i|
  create_plate(i + 1, variables, constants_accessor)
end
```

stevieing: I have done it slightly differently but pretty similar.

PR #392, opened 2020-05-18
Gpl 403 seed ont tag set

### LINKED ISSUE #221: GPL-403 - As a user I need a tag set for ONT
**User story**
As an ONT user we need to have a tag set available to add to samples when creating libraries. This will be 96 tags

Karen Oliver / Craig Corton

**Acceptance criteria**
1. When a library is created from a plate the tags should be available so these can be automatically added to the samples.

**Additional context**
What fields are required?

Chris-Friend:
Closes #221

* Update the tag set name constant used for Ont Covid
* Update the seeding of the 96 tag tag set used for Ont Covid using the oligo generated by sequencescape


Chris-Friend:
> I think you might have misunderstood the story maybe I didn't explain it correctly. We need to pull the actual tags from sequencescape which would be different in different environments i.e send a request and create the tags based on the response.

I'll add that as a requirement to the task for clarity

lib/tasks/create_tags.rake
```diff
@@ -55,19 +55,106 @@ namespace :tags do
...
-    constants_accessor = Pipelines::ConstantsAccessor.new(Pipelines.ont.covid)
-    tag_set = TagSet.create!(name: constants_accessor.pcr_tag_set_name, uuid: SecureRandom.uuid)
-    puts '-> Tag Set successfully created'
-    oligo = +'ACGTACGTACGTACGT'
-    (1..96).each do |tag_index|
-      padded_tag_number = format('%<tag_number>02i', { tag_number: tag_index })
-      Tag.create!(oligo: oligo,
-                  group_id: "ont_96_tag_#{padded_tag_number}",
-                  tag_set_id: tag_set.id)
-      oligo.increment_oligo!
```
sdjm-tessella, line 68: I will hate to see it go, but if we don't need to increment the oligo any more, the extension methods at the top of this file can be thrown away 😞

Chris-Friend: We can keep this. I'm adding a new rake task to seed the same tag set using data fetched from sequencescape

Chris-Friend:
@stevieing I think this may be more of what you were thinking: a rake task that pulls the ont tags in

PR #396, opened 2020-05-20
Added multi database support for sqlite and mysql.

stevieing:
Closes #

* This allows you to use any database you like by adding to database.yml 
* You can set the db in environment.
* Set to mysql in production
* can anyone see any problems


pjvv:
Looks fine to me @stevieing but my preference is to keep development as close to production as possible i.e. mysql :P

stevieing:
Need to fix travis integration.

PR #397, opened 2020-05-20
Added study uuid for uat and production

stevieing:
Closes #

*
*
* ...

### DIFF
--- a/config/pipelines.yml
+++ b/config/pipelines.yml
@@ -237,7 +237,7 @@ test:
         name: *pcr_tag_set_name
         hostname: 'localhost:3000'
       message: *ont_covid_message
-
+        
 uat:
   <<: *default
   ont:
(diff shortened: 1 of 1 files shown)


config/pipelines.yml
```diff
@@ -237,15 +237,19 @@ test:
         name: *pcr_tag_set_name
         hostname: 'localhost:3000'
       message: *ont_covid_message
-
+
```
sdjm-tessella, line 240: This could do with being removed, but isn't actually harming anything. My editor auto-removes this stuff on save.

config/pipelines.yml
```diff
@@ -237,15 +237,19 @@ test:
...
   <<: *default
   ont:
     covid:
       pcr_tag_set:
         name: *pcr_tag_set_name
         hostname: 'uat.sequencescape.psd.sanger.ac.uk'
-      message: *ont_covid_message
+      message: 
+        <<: *ont_covid_message
+        study_uuid:
```
sdjm-tessella, line 250: This needs to go one level deeper in the `fields:` I think

config/pipelines.yml
```diff
@@ -237,15 +237,19 @@ test:
...
         hostname: 'localhost:3000'
       message: *ont_covid_message
-
+        
 uat:
   <<: *default
   ont:
     covid:
       pcr_tag_set:
         name: *pcr_tag_set_name
```
sdjm-tessella, line 246: This references the value in the main block above, but it probably shouldn't be there.  Maybe this should be defined for the first time in the `development:` settings and then referenced for the other 3 environments from there.

PR #398, opened 2020-05-20
Update study UUID

### LINKED ISSUE #394: GPL-470 Condense ONT warehouse messages down to one large nested message
**User story**
When ONT runs are created or updated via the GraphQL v2 API, separate messages are sent to the unified warehouse (MLWH) for every sample contained in each library added to flowcells in the run.  Where there could be 5 flow cells in a run and each could contain a library with 96 samples, this would be 480 messages published for a single run.

Pacbio runs generate a single message with an array of wells and an array of samples in those wells.  ONT runs should do something similar where we nest libraries for a run as an array and each of those nest the samples in that library.

Steve Inglis

**Acceptance criteria**
1. The message specification for ONT runs should allow for nested arrays of libraries and samples.
2. The single message should be correctly generated for a run when it is created or updated.

sdjm-tessella:
Closes #394

* Add new separate environment constants YAML file for environment specific values.
* Set up nested message creation for Ont Runs (one message per entire run).
* Update tests for new expectations from ConstantsAccessor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@
 config/database.yml
 config/mailer.yml
 config/bunny.yml
+config/env_constants.yml
 
 release.tar.gz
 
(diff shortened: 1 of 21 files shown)


Chris-Friend:
Just added changes to make sure the rake tasks use the new ConstantsAccessor class methods

PR #399, opened 2020-05-20
Improve run lookup performance

sdjm-tessella:
Closes #

* Very significantly improve the performance of lookups of the run.  Turns thousands of database requests for message preparation into only 7 requests.
  * The run by ID
  * The associated flowcells for the run
  * The associated libraries for the flowcells
  * The associated requests for the libraries
  * The associated tag_taggables for the requests
  * The associated tags for the tag_taggables
  * The associated tag set for the tags

### DIFF
--- a/config/pipelines.yml
+++ b/config/pipelines.yml
@@ -181,7 +181,7 @@ default: &default
           value: request.sample_name
           populate_on_row_type: :sample
   ont:
-    message: &ont_covid_message
+    message:
       lims: *lims
       key: oseq_flowcell_run
       fields:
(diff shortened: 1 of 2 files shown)


config/pipelines.yml
```diff
@@ -181,77 +181,51 @@ default: &default
...
-
-uat:
-  <<: *default
-  ont:
-    covid:
-      pcr_tag_set:
-        name: *pcr_tag_set_name
-        hostname: 'uat.sequencescape.psd.sanger.ac.uk'
-      message: *ont_covid_message
+    message: &ont_covid_message
```
stevieing, line 184: Don't think we need &ont_covid_message anymore?

PR #401, opened 2020-05-21
Rename endpoints for ONT mutations

sdjm-tessella:

* Update endpoints for GraphQL to refer to ONT instead of Covid.
* Note that this also changes the response to the new mutation name.
* Also note that although a new argument has been added for the study type, you don't need to pass it a value at this time.  COVID is assumed.


lib/graphql_schema.json
```diff
@@ -21,8 +21,8 @@
         },
         {
           "kind": "INPUT_OBJECT",
-          "name": "CreateCovidLibrariesMutationInput",
-          "description": "Autogenerated input type of CreateCovidLibrariesMutation",
+          "name": "CreateOntLibrariesMutationInput",
```
sdjm-tessella, line 24: This file is autogenerated so feel free to ignore it

lib/graphql_schema.graphql
```diff
@@ -4,9 +4,9 @@ schema {
 }
 
 """
-Autogenerated input type of CreateCovidLibrariesMutation
+Autogenerated input type of CreateOntLibrariesMutation
```
sdjm-tessella, line 7: This file is autogenerated so feel free to ignore it

PR #404, opened 2020-05-22
Gpl 445 performance improvements

### LINKED ISSUE #355: GPL-445 Investigate performance issues around creating plates
**Description**
Having created rake tasks to generate plates with 96 tagged samples in each of 96 wells, it has become clear that performance in creating plates has become an issue.  We need to investigate the underlying cause of this since the requests take too long right now and should, ideally, be under 30 seconds.

**Users**
Stephen Inglis

Stuart McHattie and Chris Friend developed the factories related to plate/well/sample creation.

A partial investigation has already been done on this issue.  Some optimisations were put in place.  The results of which are as below:

Before any optimisation: 300+ seconds to create the pooling scenario 3 plate.
Optimising tag fetching for samples in wells: 293 seconds.
Removing duplicate validation at the start of factory save methods: 229 seconds.
Running the database as MySQL: 179 seconds.

These are all based on performance of a development MacBook Pro.  UAT ran these queries quicker, but still quite slowly.

The queries being executed in MySQL during the plate creation [is attached](https://github.com/sanger/traction-service/files/4586276/development_mysql.log.zip).  This file may be small as a .zip but inflates to 29MB of SQL statements.  At the bottom of these logs, it shows that ActiveRecord is taking up around 32 seconds, so the rest must be processing we are doing.  There are also 141.5 million allocations made.  The number of objects going into ActiveRecord should be 1 plate, 96 wells, 9,216 samples, 9,216 join objects, and 9,216 tag joins.  So the numbers are significantly lower than the allocations being made.

A good approach to this investigation could be to remove branches of the factory tree.  i.e. what happens if we don't apply tags to the samples?  What about if we look up the tags but don't create the samples?  If we can narrow the time hog to the object that's causing it, we can likely optimise the process.

Chris-Friend:
Closes #355 

* Give each active record in the Ont pipeline a `includes_args` class method which returns a list of arguments to call in the `includes` class method.
* Allow specifying arguments to remove from this list to avoid circular dependencies (e.g. when drilling down from a run to a flowcell, you don't want the flowcell's args to again include the run)
* Where needed add a `resolved_query` method that returns a queryable object with all dependencies resolved
* Use the `resolved_query` method for GraphQL queries and mutations

### DIFF
--- a/app/graphql/mutations/update_covid_run_mutation.rb
+++ b/app/graphql/mutations/update_covid_run_mutation.rb
@@ -21,13 +21,11 @@ module Mutations
       errors = update_run_flowcells(run, properties.flowcells)
       run_was_updated ||= errors.nil?
 
-      if errors&.any?
-        { run: nil, errors: errors }
-      else
-        resolved_run = Ont::Run.resolved_query.find_by(id: run.id)
-        send_messages(resolved_run: resolved_run) if run_was_updated
-        { run: resolved_run, errors: [] }
-      end
+      return { run: nil, errors: errors } if errors&.any?
+
+      resolved_run = Ont::Run.resolved_query.find_by(id: run.id)
+      send_messages(resolved_run: resolved_run) if run_was_updated
+      { run: resolved_run, errors: [] }
     end
 
     private
--- a/app/graphql/mutations/update_covid_run_mutation.rb
+++ b/app/graphql/mutations/update_covid_run_mutation.rb
@@ -24,8 +24,9 @@ module Mutations
       if errors&.any?
         { run: nil, errors: errors }
       else
-        send_messages(run: run) if run_was_updated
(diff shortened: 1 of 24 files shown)


app/graphql/mutations/update_covid_run_mutation.rb
```diff
@@ -21,12 +21,11 @@ def resolve(id:, properties:)
       errors = update_run_flowcells(run, properties.flowcells)
       run_was_updated ||= errors.nil?
 
-      if errors&.any?
-        { run: nil, errors: errors }
-      else
-        send_messages(run: run) if run_was_updated
-        { run: run, errors: [] }
-      end
+      return { run: nil, errors: errors } if errors&.any?
```
sdjm-tessella, line 24: This is a better layout for the code :+1:

PR #411, opened 2020-05-29
Gpl 445 factory perf improv

### LINKED ISSUE #355: GPL-445 Investigate performance issues around creating plates

Chris-Friend:
Closes #355 

* Add a `PlateWithSamplesFactory` which wraps around the plate factory to allow bulk insertions in a single transaction. This considerably improves plate insert performance, e.g. from about 230 to 5.5 seconds for a plate with 96 samples per well
* Keep the validation in the existing factories: still return errors if the input attributes are not as expected
* Add a new TagSetService in order that tags can be pulled into memory once per plate insert (rather than once per request)
* Update the `CreatePlateWithSamplesMutation` graph ql mutation to instead accept an array of samples


app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,153 @@
...
+    end
+
+    def process
+      @plate_factory = Ont::PlateFactory.new(attributes)
+    end
+
+    def save(**options)
+      return false unless options[:validate] == false || valid?
+
+      @serialised_plate_data = plate_factory.bulk_insert_serialise(self, validate: false)
```
sdjm-tessella, line 27: More an observation than something that absolutely needs updating, but the naming of this method is probably less relevant now it's not obvious this class is the bulk inserter.  Could probably simplify it to just `serialize`.

app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+# Ont namespace
+module Ont
+  # PlateWithSamplesFactory
+  # A factory for bulk inserting a plate and all of its dependents
+
+  # rubocop:disable Metrics/ClassLength
+  class PlateWithSamplesFactory
```
sdjm-tessella, line 9: It's too much faff for now, I think, but it would be nice if in the future this were split into two classes with different responsibilities.  Something like a `Serializer` which deals with serializing the hashes and then a `DatabaseInserter` that is responsible for only bulk inserting the data from those hashes.

app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,153 @@
...
+      Plate.insert_all!([plate_data])
+      Plate.find_by!(barcode: plate_data[:barcode])
+    end
+
+    def insert_wells_and_ont_requests(plate_id)
+      requests_data = []
+      wells_data = serialised_plate_data[:well_data].map do |well_data|
+        requests_data.concat(well_data[:request_data].map { |req_data| req_data[:ont_request] })
+        well_data[:well].merge({ plate_id: plate_id })
+      end
```
sdjm-tessella, line 104: Rather than creating wells with the map and appending requests as a separate action inside the map (which is a small abuse of the `map` function) it would probably have been better to output arrays of two items.  The first is the well data and the second is an array of requests for that well.  So you'd get something like:
```
parsed_data = [
  [well1_data, [well1_req1, well1_req2]],
  [well2_data, [well2_req1, well2_req2]]
]
```
Then you `map` the first values out to `wells_data` and `flat_map` the `requests_data` out from the second values.  Something like:
```
wells_data = parsed_data.map(&:first)
requests_data = parsed_data.flat_map(&:last)
```

app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,153 @@
...
+
+    def tag_taggagble(taggable_id, tag_id)
+      {
+        taggable_type: 'Ont::Request',
+        taggable_id: taggable_id,
+        tag_id: tag_id
+      }.merge(timestamps)
+    end
+  end
+  # rubocop:enable Metrics/ClassLength
```
sdjm-tessella, line 152: This would likely go away if this were two classes.

app/services/tag_set_service.rb
```diff
@@ -0,0 +1,21 @@
...
+# A service for fetching whole tag sets into memory
+class TagSetService
+  attr_reader :loaded_tag_sets
+
+  def initialize
+    @loaded_tag_sets = {}
+  end
+
+  def load_tag_set(tag_set_name)
+    return if loaded_tag_sets.key?(tag_set_name)
```
sdjm-tessella, line 13: I would expect this would be an error in the workflow to receive anything but the correct key.  Maybe it should cause an exception rather than just overwriting the loaded tag set.

spec/models/ont/plate_with_samples_factory_spec.rb
```diff
@@ -0,0 +1,357 @@
...
+          expect(ContainerMaterial.all.count).to eq(0)
+        end
+  
+        it 'does not insert any ont_requests' do
+          expect(Ont::Request.all.count).to eq(0)
+        end
+  
+        it 'does not insert any tag_taggables' do
+          expect(TagTaggable.all.count).to eq(0)
+        end
```
sdjm-tessella, line 315: It's good to see this being tested, though I'd like to see that all these database assertions are true, no matter which insert failed.  It might be that if you let the plate insert, and the wells, and the requests, but not the tag_taggables, that you have half the items in the database as this isn't asserted anywhere.

app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,162 @@
+# frozen_string_literal: true
+
+# Ont namespace
+module Ont
+  # PlateWithSamplesFactory
+  # A factory for bulk inserting a plate and all of its dependents
+
+  # rubocop:disable Metrics/ClassLength
```
stevieing, line 8: Probably better to fix this rather than circumvent it. We should only be using rubocop disable for legacy code.

app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    def get_request_ids_by_uuid(request_uuids)
+      Ont::Request.where(uuid: request_uuids).map { |req| [req.uuid, req.id] }.to_h
+    end
+
+    def get_well_ids_by_position(wells)
+      wells.map { |well| [well.position, well.id] }.to_h
+    end
+
+    # rubocop:disable Metrics/AbcSize
```
stevieing, line 124: Again. Needs to be fixed.

app/models/ont/plate_with_samples_factory.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    def tag_taggable(taggable_id, tag_id)
+      {
+        taggable_type: 'Ont::Request',
+        taggable_id: taggable_id,
+        tag_id: tag_id
+      }.merge(timestamps)
+    end
+  end
+  # rubocop:enable Metrics/ClassLength
```
stevieing, line 161: If the class is too big then we need to abstract.

PR #417, opened 2020-06-02
removed run name from pacbio request params

BenTopping:

*Remove name from pacbio run parameters since it is now auto generated


spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -155,7 +154,7 @@
       it 'updates a run' do
         patch v1_pacbio_run_path(@run), params: body, headers: json_api_headers
         @run.reload
-        expect(@run.name).to eq "an updated name"
+        expect(@run.name).to eq "TRACTION-RUN-1"
```
harrietc52, line 157: can remove this line

PR #432, opened 2020-06-15
Generate sample sheet broken in uat

### LINKED ISSUE #428: BUG - Generate Sample Sheet broken in UAT
Expect sample sheet to be able to download in UAT, but broken

stevieing:
Closes #428 

*
*
* ...


stevieing:
> Looks good, how is `all_wells_have_libraries` in the run response to be used in the UI?

If you look at the corresponding pull request for the ui that will show you. #432 in traction-ui.

It is a bit of a hack. A better way would be to have a state machine for run which means we could show buttons based on its state. We might have to do this as part of the event work.

harrietc52:
UI https://github.com/sanger/traction-ui/pull/480

PR #455, opened 2020-07-09
[WIP] Template prep kit box barcode fix (service)

BenTopping:
Closes #

-Failing lint currently

Back end changes for moving template_prep_kit_box_barcode to library from run
Coupled with ui PR: https://github.com/sanger/traction-ui/pull/495
* Changes CSV generator to use library template prep kit box barcode
* Removed template prep kit box barcode from run model
* Replaced library kit barcode with template prep kit box barcode for libraries

This PR contains migrations


harrietc52:
Just to bear in mind data migration :) might need a rake task

stevieing:
> Just to bear in mind data migration :) might need a rake task

I don't think it should be necessary. We will just need to check that the template prep kit box barcodes are the same as the library kit barcode.

app/csv_generator/csv_generator.rb
```diff
@@ -59,6 +64,13 @@ def csv_data(obj, is_well_header_row, sample_names)
     end
   end
 
+  def check_template_prep_kit_box_barcode(well_libraries)
```
stevieing, line 67: Might be better to put this method in well. That way you don't need any logic in the csv generator. Then you can just have template_prep_kit_box_barcode in pipelines.yml

config/pipelines.yml
```diff
@@ -129,9 +129,7 @@ default: &default
           value: insert_size
           populate_on_row_type: :well
         Template Prep Kit Box Barcode:
```
stevieing, line 131: So based on the changes above this would become
```
Template Prep Kit Box Barcode:
 type: :model
 value: template_prep_kit_box_barcode
 populate_on_row_type: :well
```

spec/csv_generator/csv_generator_spec.rb
```diff
@@ -44,7 +44,7 @@
         well1.sample_names,
         well1.movie_time.to_s,
         well1.insert_size.to_s,
-        well1.plate.run.template_prep_kit_box_barcode,
+        'LK1234567',
```
stevieing, line 47: I think you will now need to change these to well.template_prep_kit_box_barcode

PR #469, opened 2020-08-11
Added ability to populate well position for all rows to csv generator.

### LINKED ISSUE #462: GPL-589 Sample Well needs to be provided on all rows of the csv file.
**User story**
As Craig, I would like the Sample Well information to be provided on each row in in the sample sheet CSV

Harriet
Steve
Craig
Karen

**Acceptance criteria**
- [ ] SMRT sample sheet can be uploaded
- [ ] a well with multiple libraries works. Both rows show well position
- [ ] a well with one library. The row shows position

stevieing:
Closes #462 

*
*
* ...


spec/models/pacbio/request_library_spec.rb
```diff
@@ -44,4 +44,12 @@
     end
   end
 
+  context 'collection?' do
+    let(:well)                { create(:pacbio_well) }
+
+    it 'will always be true' do
+      expect(well).to be_collection
+    end
+  end
```
KatyTaylor, line 53: Should we be creating a pacbio request library here, rather than a well, and shouldn't collection? be false rather than true?

spec/models/pacbio/well_spec.rb
```diff
@@ -152,7 +152,15 @@
...
-      expect(well.template_prep_kit_box_barcode).to eq 'Lxxxxx100938900123199'
+      expect(well.template_prep_kit_box_barcode).to eq Pacbio::Well::GENERIC_KIT_BARCODE
+    end
+  end
+
+  context 'collection?' do
+    let(:well)                { create(:pacbio_well) }
+
+    it 'will always be false' do
+      expect(well).to be_collection
```
KatyTaylor, line 163: think `will always be false` should say `will always be true` ?

app/csv_generator/csv_generator.rb
```diff
@@ -37,32 +37,55 @@ def csv_headers
...
+  # type: :model
+  # value: position_leading_zero
+  # populate:
+  #   for:
+  #     - :well
+  #     - :sample
+  #   with: :well+
+  # means that sample well needs to be populated for both samples and wells
+  # but needs to use the well method
+  # hopefully that is enough of an explanation!
```
KatyTaylor, line 82: nice explanation 👍  I notice you've got little '+' signs at the end of lines 69 and 79 - can you get rid of them, or are they are a github artefact?

spec/models/pacbio/request_library_spec.rb
stevieing: Yes. Just copied it from the other one.

app/csv_generator/csv_generator.rb
stevieing: The plus signs are for rdoc to signify a code block.

app/csv_generator/csv_generator.rb
```diff
@@ -37,32 +37,55 @@ def csv_headers
...
+  def populate_column(options = {})
+    populate = options[:column_options][:populate]
+    return '' unless populate[:for].include?(options[:row_type])
+
+    obj = populate[:with] == :row_type ? options[options[:row_type]] : options[populate[:with]]
+    instance_value(obj, options[:column_options])
```
KatyTaylor, line 88: I would personally make this a bit more verbose, as I'm finding it a bit hard to get my head around despite the excellent explanation in the comment!

I think the main confusion is knowing what the structure of the `options` argument is, without having to look back at the previous methods. Maybe it would be better to pass in the `sample`, `well`, `row_type` and `column_options` arguments explicitly?

spec/csv_generator/csv_generator_spec.rb
```diff
@@ -94,8 +94,8 @@
         '',
         '',
         'false',
+        well1.position,
```
KatyTaylor, line 97: What are these changes for? Were they in the wrong order before?

app/csv_generator/csv_generator.rb
stevieing: I can add a comment that explains the arguments as a compromise. I did look at passing in those arguments explicitly but then you have to either do send or instance eval which will cause compexlity elsewhere. I thought that the new structure of the yaml file explained those options in a sematic way.

app/csv_generator/csv_generator.rb
KatyTaylor: Ah good to know, thanks

spec/csv_generator/csv_generator_spec.rb
stevieing: Samples now need to have well position and I have removed sample names as it was no longer required.

KatyTaylor: Ah got it thanks

PR #494, opened 2020-10-26
Gpl 686 modify library tags

### LINKED ISSUE #490: GPL-685 When I update a library that is part of a run this information should be disseminated to the warehouse
**User story**
When I update a library that is part of a run this information should be disseminated to the warehouse 

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
- [ ] If I change a tag on a library that is part of a run this change should be reflected in the warehouse.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #490 

*Added new request library endpoint
*request library can now be updated
* ...

### DIFF
--- a/spec/requests/v1/pacbio/request_library_spec.rb
+++ b/spec/requests/v1/pacbio/request_library_spec.rb
@@ -1,0 +1,37 @@
+require "rails_helper"
+
+RSpec.describe 'RequestLibraryController', type: :request do
+
+    context '#update' do
+        let(:tag) { create(:tag) }
+        let(:request_library) { create(:pacbio_request_library) }
+
+        context 'on success' do
+            let(:body) do
+                {
+                data: {
+                    type: "tags",
+                    id: request_library.id,
+                    attributes: {
+                    tag_id: tag.id
+                    }
+                }
+                }.to_json
+            end
+
+            it 'has a ok status' do
+                patch v1_pacbio_request_library_path(request_library), params: body, headers: json_api_headers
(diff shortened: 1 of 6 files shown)


BenTopping:
Sample ID is no longer needed, will remove tomorrow

spec/requests/v1/pacbio/request_library_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+            end
+
+            it 'returns the correct attributes' do
+                patch v1_pacbio_request_library_path(request_library), params: body, headers: json_api_headers
+                json = ActiveSupport::JSON.decode(response.body)
+                expect(json['data']['id']).to eq request_library.id.to_s
+            end
+        end
+
+        #context 'on failure' cannot happen since there is no validation as you cannot send a requestLibrary without a tag_id
```
harrietc52, line 34: Still worth adding the check to make sure it is handled in the back end. Server code shouldn't rely on the client codes validation, it should do some itself. In the controlled I think the params `permit` method would fail if it wasnt present so maybe adding a test to check that.

harrietc52: Just check the formatting in this file too :) curly braces are on odd lines

harrietc52 (commented):
Looks great! Small comment about a test

harrietc52 (approved):

PR #530, opened 2021-02-09
Migrating off travis

BenTopping:

*Ignore multiple merges in commit history, was due to testing build by merging into develop on forked branch
*Automated releases built for master and develop upon merge
*Rpsec and rubocop tests for push/pull requests

Questions:
- Should i add a workflow for creating a release manually?


JamesGlover (approved):

JamesGlover:
I'm not really familiar with the way the traction releases are managed, so its probably a good idea for @stevieing to also have a look at that section. In most the apps I've been migrating, releases are all added manually.

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+      uses: actions/create-release@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+      with:
+        tag_name: ${{ env.RELEASE_TAG }}
+        release_name: ${{ env.RELEASE_TAG }}
+        draft: false
+        prerelease: false
+
+    - name: Set up environment
```
JamesGlover, line 34: Hmm, actually, on reflection I wonder if it makes more sense to split this into two actions. One that makes the release, and then another that makes the build. If the latter is triggered on
```
on:
  release:
    types: published
```
Then I *think* one would automatically chain into the other, but you'd leave scope to create a manual release as well.

BenTopping: yes that sounds like a good idea ill look into that

JamesGlover: Haha, I've just gone to pick up quanthub, and it looks like it follows a similar release pattern. So I'm going to be referencing what you've done here a bit. Let you know if I find out anything that might be useful.

JamesGlover: So used `${GITHUB_REF#refs/heads/}` to extract the branch name. Not hugely significant in this case, where we're just matching develop and master, but it means it will correctly handle branches with / in the name. Mostly made the change in my version just in case someone tries to nab the code for some other purpose.

JamesGlover: I've also used:
`prerelease: ${{ github.ref == 'refs/heads/devel' }}`
To automatically flag devel releases as pre-releases

stevieing (approved):

PR #566, opened 2021-05-14
Gpl 990 import samples from plates as well as tubes

### LINKED ISSUE #555: GPL-990 - As a long read user (Craig) I need to be able to import samples from plates as well as tubes
**User story**
As a long read user (Craig) I need to be able to import samples from plates as well as tubes

e.g. John S (don't include surnames in public repos)

**Acceptance criteria**
- [ ] add a list of acceptance criteria here
- [ ] ...

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

stevieing:
Closes #555 

* moved the pacbio sequencing plate and well resources to be namespaced under runs
* created a plate creator model to build a plate with wells and samples from a plate that exists in Sequencescape
* created an endpoint to return all of the plates that are pacbio (/v1/pacbio/plates)
* created an endpoint to create a plate for pacbio (/v1/pacbio/plates [POST])

Things to note:
- I have not created an update or delete endpoint for plates (they should be read only)

problems with resources:
- because the well is a container the resource returns a barcode so it probably makes sense to tidy up the resources to be a bit more specific
- It would be better if the well resource returns samples rather than materials

Things to sort:
- we need the study id for each sample
- do we need to populate the source barcode


app/models/pacbio/plate_creator.rb
```diff
@@ -0,0 +1,237 @@
...
+      def initialize(attributes = {})
+        @plate = ::Plate.new(attributes.except(:wells))
+        @well_wrappers = attributes[:wells].try(:collect) do |well|
+          WellWrapper.new(well.merge(plate: plate))
+        end
+      end
+
+      def save!
+        plate.save!
+        raise ActiveRecord::RecordInvalid if well_wrappers.nil? || well_wrappers.try(:empty?)
```
BenTopping, line 91: could this be changed to just 'raise ActiveRecord::RecordInvalid unless valid?' or if avoiding adding error messaging does well_wrappers.blank? accomplish the same thing

spec/requests/v1/pacbio/plates_spec.rb
```diff
@@ -85,110 +103,15 @@
...
-      it 'has a ok status' do
-        patch v1_pacbio_plate_path(plate), params: body, headers: json_api_headers
-        expect(response).to have_http_status(:ok)
-      end
-
-      it 'updates a plate' do
-        patch v1_pacbio_plate_path(plate), params: body, headers: json_api_headers
-        plate.reload
-        expect(plate.run).to eq new_run
+        expect(json['data']['errors']).to be_present
```
BenTopping, line 112: Make sure they are the correct errors, not just that they exist

spec/requests/v1/pacbio/runs/plates_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+      expect(json['data'][0]['attributes']['pacbio_run_id']).to eq(plate1.pacbio_run_id)
+      expect(json['data'][1]['attributes']['pacbio_run_id']).to eq(plate2.pacbio_run_id)
+
+      wells = json['included']
+      expect(wells.length).to eq(5)
+
+      well = wells[0]['attributes']
+      plate_well = plate2.wells.first
+      expect(well['insert_size']).to eq(plate_well.insert_size)
```
BenTopping, line 33: could check the entire contents of well against the contents of plate_well

BenTopping:
Is there a plan to give all existing plates a pipeline aswell as in do they need a pipeline value added in the schema? I have noticed the endpoint /v1/pacbio/plates isnt returning any plates for me because the plates don't have a pipeline value?

spec/requests/v1/pacbio/plates_spec.rb
stevieing: > Make sure they are the correct errors, not just that they exist

I already do that in the plate creator model.

spec/requests/v1/pacbio/runs/plates_spec.rb
stevieing: > could check the entire contents of well against the contents of plate_well

Again I think I have already tested that elsewhere.

PR #576, opened 2021-05-28
Library source name

JamesGlover:
Closes #

*
*
* ...


app/resources/v1/pacbio/run_resource.rb
```diff
@@ -20,6 +20,22 @@ def created_at
       def all_wells_have_libraries
         @model.all_wells_have_libraries?
       end
+
```
stevieing, line 23: urgh. Let's leave it like this for now. I would be more tempted to fix it so that we only have a single model for plate.

stevieing (approved):

PR #585, opened 2021-06-10
setting default cost code value on pacbio request creation

BenTopping:
When importing samples into traction pacbio, the created requests should have a default cost code if one isnt provided.

NB will need to change front end requests builds if sending cost_code as null and not undefined as rails attribute default only works if no value exists.

*
*
* ...


JamesGlover (approved):

lib/tasks/pacbio_data.rake
```diff
@@ -14,23 +14,40 @@ namespace :pacbio_data do
...
+    end
+    puts '-> Pacbio libraries successfully created'
+
+    puts '-> Creating pacbio runs...'
+    (1..5).each do |i|
       run = Pacbio::Run.create!(name: "Run#{i}", binding_kit_box_barcode: "BKB#{i}",
                                 sequencing_kit_box_barcode: "SKB#{i}", dna_control_complex_box_barcode: "DCCB#{i}")
       plate = Pacbio::Plate.create!(run: run)
-      Pacbio::Well.create!(plate: plate, libraries: [library], movie_time: 20, insert_size: 10, on_plate_loading_concentration: 1,
+      Pacbio::Well.create!(plate: plate, libraries: [Pacbio::Library.order('RANDOM()').first], movie_time: 20, insert_size: 10, on_plate_loading_concentration: 1,
```
BenTopping, line 50: may need to be 'RAND()' for mysql, im still on sqlite

PR #610, opened 2021-07-15
Gpl 878 10 well library removal

BenTopping:
Closes #

**This branch is branched off GPL-878-10-add-pool-to-well-in-pacbio-run **
 
TODO:
Check sample sheet generation is working as expected (go to pacbio->runs->generate sample sheet button)
Check resources (Had a quick go at building new resources for the runs work but wasnt sure of the data needed at this time
Check wellPool validation - Currently there is no validation for wellPools, previous library validations comment in well_factory model


app/controllers/v1/pacbio/runs/wells_controller.rb
```diff
@@ -63,22 +63,12 @@ def param_names
           well_param_names(p1)
```
JamesGlover, line 63: Your changes here are fine, but I think there's a bit of scope for overall improvement here. A few things I've noticed:

1) A lot of code repetition, we seem to be permitting things several times
2) We're passing round various flavours of 'params' despite it being an object variable
3) We've a few points where we're chaining hash [] methods, and then assuming we're getting a particular
    data type back. eg. `params.require(:relationships)[:libraries][:data].map`

app/models/pacbio/well_factory.rb
```diff
@@ -112,94 +102,24 @@ def save
         return false unless valid?
 
         well.save
-        return false unless libraries.save && pools.save
+        return false unless pools.save
 
         true
```
JamesGlover, line 107: Here `well.save` returns false if save fails (unlikely because we just validated, but still possible) and yet we'll ignore that and head on to persist the pool.

This could just become `well.save && pool.save`. That way we'd only save the pool if the well saved, and would return false if either failed. (And true if bot succeeded)

app/resources/v1/pacbio/runs/tube_resource.rb
```diff
@@ -1,13 +0,0 @@
...
-  module Pacbio
-    module Runs
-      # TubeResource
-      class TubeResource < JSONAPI::Resource
-        model_name 'Tube'
-        attributes :barcode
-      end
-    end
-  end
-end
```
JamesGlover, line 13: Wont we still want this on pool?

app/resources/v1/pacbio/runs/library_resource.rb
```diff
@@ -0,0 +1,13 @@
...
+  module Pacbio
+    module Runs
+      class LibraryResource < JSONAPI::Resource
+        model_name 'Pacbio::Library'
+        attributes :state, :volume, :concentration, :template_prep_kit_box_barcode,
+                   :fragment_size, :created_at, :deactivated_at, :source_identifier
+      end
+    end
+  end
+end
```
JamesGlover, line 13: Why do we need a different library resource under the run namespace? Can't we just continue to use `app/resources/v1/pacbio/library_resource.rb`

app/resources/v1/pacbio/runs/well_pool_resource.rb
```diff
@@ -0,0 +1,14 @@
...
+    module Runs
+      # WellPoolResource
+      class WellPoolResource < JSONAPI::Resource
+        model_name 'Pacbio::WellPool'
+
+        has_one :pool
+      end
+    end
+  end
+end
```
JamesGlover, line 14: I know we exposed Well:Library before, but I'm not sure there's much value in exposing what is essentially a join-table. We should probably hide this from the API as an implementation decision.

app/resources/v1/pacbio/runs/pool_resource.rb
```diff
@@ -0,0 +1,14 @@
...
+    module Runs
+      # PoolResource
+      class PoolResource < JSONAPI::Resource
+        model_name 'Pacbio::Pool'
+
+        has_many :libraries
+      end
+    end
+  end
+end
```
JamesGlover, line 14: Similar question to Library resource, what's the reasoning behind the separate run namespaced resource?

app/resources/v1/pacbio/runs/well_resource.rb
```diff
@@ -11,7 +11,7 @@ class WellResource < JSONAPI::Resource
                    :row, :column, :pacbio_plate_id, :comment, :generate_hifi,
                    :position, :pre_extension_time, :ccs_analysis_output
 
-        has_many :libraries, class_name: 'WellLibrary', relation_name: :well_libraries
+        has_many :pools, class_name: 'WellPool', relation_name: :well_pools
```
JamesGlover, line 14: We probably should expose the pools directly here, rather than the join model.

spec/csv_generator/csv_generator_spec.rb
```diff
@@ -4,16 +4,16 @@
...
     let(:plate)   { create(:pacbio_plate, wells: [well1, well2]) }
     let(:run)     { create(:pacbio_run, plate: plate) }
     let(:csv)     { ::CsvGenerator.new(run: run, configuration: Pipelines.pacbio.sample_sheet) }
 
     it 'check validity' do
       well = create(:pacbio_well)
-      well.libraries = create_list(:pacbio_library, 5)
-      expect(true).to be_truthy
+      well.pools << create_list(:pacbio_pool, 5)
+      expect(well).to be_valid
```
JamesGlover, line 16: 😆

app/models/pacbio/well_factory.rb
```diff
@@ -237,6 +157,24 @@ def destroy_pools
...
+      #   return unless all_tags.empty? || all_tags.any?(nil)
+
+      #   errors.add(:tags, 'are missing from the libraries')
+      # end
+
+      # def check_tags_uniq
+      #   return if all_tags.length == all_tags.uniq.length
+
+      #   errors.add(:tags, "are not unique within the libraries for well #{well.position}")
+      # end
```
JamesGlover, line 171: So I think we'll need these two....

app/models/pacbio/well_factory.rb
```diff
@@ -237,6 +157,24 @@ def destroy_pools
...
+
+      # def check_libraries_max
+      #   return if libraries.length <= 16
+
+      #   errors.add(:libraries, "There are more than 16 libraries in well #{well.position}")
+      # end
```
JamesGlover, line 177: ... but not this one

JamesGlover (commented):
- Aware that a few of the comments I've made aren't really a result of your changes, feel free to keep them low priority.
- The model stuff seems pretty solid
- Commented on the validations, but aware that's in the todo bit still

I'm not sure about the approach on resources, but you said you weren't sure about that. Mostly its a case of:
1) Avoid exposing join-tables. I know we were doing this before, but it leaks implementation details, and has no value to the API itself. 
2) You had a few resources duplicated under the 'run' namespace, where it would be better if we could just use the Pacbio namespaced version. I know @stevieing ran into a bit of trouble trying to get this to work for tags at one point, so it might be you had similar troubles.

app/models/pacbio/well_factory.rb
JamesGlover: Things not quite behaving as expected here. These may not be rails models...

JamesGlover: Hmm, pools is a factory object... makes sense

PR #619, opened 2021-08-02
Fixed isoSeq sample sheet generation

BenTopping:
Closes #

*
*
* ...


stevieing (approved):
All fine with me. A bit hacky but there is nothing we can do about it. Once we add library type hopefully we can get rid of it.

app/models/concerns/sample_sheet.rb
```diff
@@ -28,11 +28,33 @@ def barcode_set
     libraries.first.tag.tag_set.uuid
   end
 
+  # Sample bio Name field - TEMPORARY: Want well row to have sample_names when isoSeq tags
+  def find_sample_name
+    if defined?(sample_names) # When row type is well
```
JamesGlover, line 33: Won't this also hide sample names with untagged wells?
I'm also a little concerned that if someone adds 'sample_names' to library, this will change the behaviour here (and as a result we'll also be dropping sample names for untagged libraries).

JamesGlover (changes requested):

app/models/concerns/sample_sheet.rb
BenTopping: This fix is a temporary fix until we are able to determine the pipeline (type) of a sample e.g isoseq. Also sample_names should never be added to library as librarys can only have one sample.
Im not sure I understand, the intended functionality is: 

On the csv well row: If any libraries in the well are not tagged or are isoseq tagged then the method check_library_tags will return false and therefore sample_names would be returned.
On the library (sample) rows: the request sample_name is returned

app/models/concerns/sample_sheet.rb
```diff
@@ -28,11 +28,33 @@ def barcode_set
...
+  # Sample is Barcoded field - TEMPORARY: Want IsoSeq samples to return false
+  def check_library_tags
+    isoseq_tag_sets = %w[IsoSeq_Primers_12_Barcodes_v1]
+    if libraries.any? do |library|
+         isoseq_tag_sets.include?(library.tag.tag_set.name) if library.tag
+       end
+      return false
+    end
+
+    libraries.all?(&:tag_id?)
```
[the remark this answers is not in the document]
BenTopping: agreed, cleans it up a bit

app/models/concerns/sample_sheet.rb
JamesGlover: Yep... that looks like it matches up.

app/models/concerns/sample_sheet.rb
```diff
@@ -45,14 +46,11 @@ def all_libraries_tagged
...
   # Sample is Barcoded field - TEMPORARY: Want IsoSeq samples to return false
   def check_library_tags
-    isoseq_tag_sets = %w[IsoSeq_Primers_12_Barcodes_v1]
-    if libraries.any? do |library|
-         isoseq_tag_sets.include?(library.tag.tag_set.name) if library.tag
-       end
-      return false
+    no_isoseq_tag_sets = libraries.none? do |library|
+      ISOSEQ_TAG_SETS.include?(library.tag.tag_set.name) if library.tag
     end
```
JamesGlover: I actually meant pull this into a separate method, but this is fine, considering its temporary (and still has the advantage of making what the checks are explicit)

JamesGlover (approved):

PR #623, opened 2021-08-05
Gpl 878 22 populate insert size from pool

BenTopping:
Closes # 

* Remove insert size from well, now retrieved from first pool in well
* 
* ...


db/migrate/20210805135643_remove_insert_size_from_pacbio_run_well.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveInsertSizeFromPacbioRunWell < ActiveRecord::Migration[6.0]
+  def change
```
BenTopping, line 2: need to think about this migration as not sure how it is going to work with legacy data, if fragment size is already populated in pools it might be fine, but then it may not be same values as previous insert_size on old runs

app/models/pacbio/well.rb
```diff
@@ -74,6 +73,11 @@ def template_prep_kit_box_barcode
       barcodes[0]
     end
 
+    def insert_size
+      pools? ? pools.first.fragment_size : ''
+      # How to handle multiple pools in a well?
```
BenTopping, line 78: waiting on response from Craig, will forward on email once i receive it

stevieing (approved):

PR #652, opened 2021-10-04
DPL-136 sample sheet well order for pacbio runs 

### LINKED ISSUE #647: DPL-136 [BUG] - When exporting a sample sheet the order of samples in the file did not match the well order
**Describe the bug**
When exporting a sample sheet the order of samples in the file did not match the well order.

**RT Ticket Number**
If applicable

**To Reproduce**
1. pacbio > runs > select run
2. download sample sheet
3. Wells in order A1, B1, C1, D1

**Expected behaviour**
Wells in order A1, B1, D1, C1

**Screenshots**

**Desktop (please complete the following information):**
* Windows or Mac?
* Browser Chrome, Firefox, Safari or other?
* Browser version (use 'About' to look up)?

**Additional context**

BenTopping:
Wells are now sorted by column for sample sheet generation in pacbio runs

Closes #647 

*
*
* ...

### DIFF
--- a/app/csv_generator/csv_generator.rb
+++ b/app/csv_generator/csv_generator.rb
@@ -14,8 +14,9 @@ class CsvGenerator
   def generate_sample_sheet
     CSV.generate do |csv|
       csv << csv_headers
+      sorted_wells = sort_wells(run.plate.wells)
 
-      run.plate.wells.each do |well|
+      sorted_wells.each do |well|
         # add well header row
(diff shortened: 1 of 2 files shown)


stevieing (changes requested):
This all works and looks good.

Looking at the current code base there is already a well sorter service and also some sorting in the plate model.

Maybe it would be better to reuse what is already there?

PR #681, opened 2021-12-22
Issue639/dpl 124 add the concept of library type to a request {c m v 4}

### LINKED ISSUE #639: DPL-124 - Add the concept of library type to a request (C=M, V=4)
**User story**
NPG need to be able to identify IsoSeq runs. To that end we need to add the concept of library type

Steve

**Acceptance criteria**
- [x] When a sample is imported it can accept library type (not required)
- [x] When a run is exported to the MLWH it will include the library type if it exists

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

JamesGlover:
Closes #639

* Add library type to flowcell message

Dependent on: https://github.com/sanger/unified_warehouse/pull/314 being released before this.

### DIFF
--- a/config/pipelines.yml
+++ b/config/pipelines.yml
@@ -102,6 +102,10 @@ default: &default
                 tag_set_name:
                   type: :model
                   value: tag.tag_set_name
+                pipeline_id_lims:
+                  type: :model
+                  value: request.library_type
+
     sample_sheet:
       columns:
         System Name:
--- a/app/controllers/v1/pacbio/plates_controller.rb
+++ b/app/controllers/v1/pacbio/plates_controller.rb
@@ -29,8 +29,7 @@ module V1
       def plates_params
         params.require(:data)['attributes'].permit(
           plates: [:barcode, { wells: [:position, { samples:
-              %i[name external_id species library_type estimate_of_gb_required
-                 number_of_smrt_cells cost_code external_study_id] }] }]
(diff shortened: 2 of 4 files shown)


config/pipelines.yml
```diff
@@ -102,6 +102,10 @@ default: &default
                 tag_set_name:
                   type: :model
                   value: tag.tag_set_name
+                pipeline_id_lims:
```
andrewsparkes, line 105: Not convinced by the name 'pipeline_id_lims' for Libary type in the messsage as it suggests that pipeline to library type is 1:1. Can it not be 'library_type'?

andrewsparkes (approved):
Looks clean to me other than my minor comment about the message field name.

config/pipelines.yml
stevieing: pipeline_id_lims is the name of the column in the warehouse so we can't change it.

stevieing (approved):

config/pipelines.yml
andrewsparkes: Ok

JamesGlover: Well the column in the warehouse doesn't exist yet, so could be renamed; but this name was picked for consistency with the iseq_flowcell table and was originally specified by NPG (I can't quite remember why they didn't want library_type).

I have highlighted the column name in an email for Carol S, so there might still be a request to change it.

PR #729, opened 2022-03-08
DPL-298  - Data integrity and tube exposure

JamesGlover:
Closes #

*
*
* ...


db/migrate/20220303085155_add_constraints_to_container_materials.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+# During development I ended up with a couple of data-integrity issues
+# due to missing validations allowing container_materials to save
+# when their associated resources were invalid. Adding belt and braces
+# approach to ensure we maintain data integrity. (Glance at production suggests
+# data is currently valid)
+#
+# A similar constraint should probably be applied to material_id and
```
stevieing, line 9: Is it worth putting in a story to fix material_id and material_type? There is a concern that once we scale up it could become a problem.

stevieing (approved):

db/migrate/20220303085155_add_constraints_to_container_materials.rb
JamesGlover: Definitely. I suspect its just a case of us needing to delete the records

PR #744, opened 2022-04-14
DPL-334 Edits (updates) not sent to MLWH

stevieing:
Closes #

*
*
* ...


app/models/pacbio/request_factory.rb
```diff
@@ -3,6 +3,115 @@
 module Pacbio
   # RequestFactory
   class RequestFactory
```
harrietc52, line 5: Could add in the module level comments from the old `Pipelines::Requestor::...` etc 
e.g.
```
    # Factory - behaviour for pipeline requests factory
    # It will take an array of request parameters and build an
    # array of request and associated samples (if they dont exist)
    # Also manages validation and saving
```

spec/factories/pacbio/plates.rb
```diff
@@ -3,7 +3,20 @@
...
+        well_factory { :pacbio_well }
+      end
+
+      wells { create_list(well_factory, well_count)}
+    end
+
+    trait :pooled do
+      transient do
+        pool_count { 1 }
+        # TODO: This needs sorting as we should be able to select the number of pools
```
harrietc52, line 17: Is this a todo for now?

spec/requests/v1/pacbio/pools_spec.rb
```diff
@@ -360,5 +363,45 @@
...
+              updated_at: '2021-08-04T14:35:47.208Z'
+            }
+          }
+        }.to_json
+      end
+      
+      it 'publishes a message' do
+        expect(Messages).to receive(:publish).with(pool.sequencing_plates, having_attributes(pipeline: 'pacbio'))
+        patch v1_pacbio_pool_path(pool), params: body, headers: json_api_headers
+        expect(response).to have_http_status(:success), response.body
```
harrietc52, line 402: Does it still publish a message if a patch has been performed but not data has actually changed? I guess that would do an update in MLWH, but shouldn't be harmful

spec/support/pipelines/requestor_controller.rb
```diff
@@ -1,171 +0,0 @@
-require 'rails_helper'
```
harrietc52, line 1: Where there any other `Pipeline::Requestor: ...` spec files to delete?

harrietc52 (approved):
Looks good, minor questions/comments

app/resources/v1/pacbio/request_resource.rb
```diff
@@ -4,11 +4,34 @@ module V1
...
+      attributes(*::Pacbio.request_attributes, :sample_name, :barcode, :sample_species,
+                 :created_at, :source_identifier)
 
       has_one :well
       has_one :plate
       has_one :tube
+
+      # When a request is updated and it is attached to a run we need
+      # to republish the messages for the run
+      after_update :publish_messages
```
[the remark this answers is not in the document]
stevieing: You are correct. That is how it works. All libraries are in a pool and we have removed the library edit from the front end. You can still access the endpoint but we think it is better not to send the message.

spec/factories/pacbio/plates.rb
stevieing: Yes. It was a bit tricky to sort and wanted to timebox the story.

spec/requests/v1/pacbio/pools_spec.rb
stevieing: Yes. It has no side effect so think it is safe. I suppose we could use the dirty active record method but I thought it was adding complexity.

spec/support/pipelines/requestor_controller.rb
stevieing: Yes. There are some left. There is still a bit of work to do but I decided to timebox the refactoring for now. I am also doing the Rails 7 upgrade as part of DPL-274 so lots of improvements happening.

PR #745, opened 2022-04-26
DPL-274 - ONT - remove existing code

stevieing:

*
*
* ...


JamesGlover:
Ahh bingo!

https://github.com/cerebris/jsonapi-resources/issues/1308

It is applying the self.records filter from `V1::Pacbio::TubeResource` which is: `::Tube.by_pipeline(:pacbio)` so we end up inadvertently filtering out out tubes.

app/resources/v1/pacbio/library_resource.rb
```diff
@@ -23,11 +27,11 @@ def self.records_for_populate(*_args)
       end
 
       def created_at
-        @model.created_at.to_s(:us)
+        @model.created_at.to_fs(:us)
```
JamesGlover, line 30: Any idea why we're calling our custom time format `:us`?
Introduced here: https://github.com/sanger/traction-service/pull/240/commits/e86453bc220535f8031da466b39a3a70163cd917

As part of this PR:
https://github.com/sanger/traction-service/pull/240

config/initializers/01_proc.rb
```diff
@@ -1,21 +0,0 @@
...
-        false
-      end
-
-      def respond_to_missing?(*)
-        super
-        true
-      end
-    end.new
-  end
-end
```
JamesGlover, line 21: I remember thinking this was a native feature I'd not seen before when I first came across it in the codebase.

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
```
JamesGlover, line 1: Given you've already opted in to the 7.0 defaults, I think we can already lose this file.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_04_07_104659) do
+ActiveRecord::Schema[6.1].define(version: 2022_04_07_104659) do
```
JamesGlover, line 13: Might be worth running `rake db:schema:dump` now to update this to the Rails 7 version, otherwise there will be lots of changes next time someone touches the schema for real.

spec/support/read_only.rb
```diff
@@ -0,0 +1,6 @@
+  # This allows us to unlock read only tables for testing
+  def set_read_only(klasses, readonly=false)
+    Array(klasses).each do |klass|
+      klass.define_method(:readonly?) { readonly }
+    end
```
JamesGlover, line 5: Might be better using mocks here as they'll get automatically torn down, removing the need for an after block, and reducing the risk of accidentally mutating the global state.

```ruby
klasses.each do |klass|
  allow_any_instance_of(Widget).to receive(:readonly?).and_return(readonly)
end
```
I think rubocop-rspec will complain, but I think its still a safer option.

app/models/ont/flowcell.rb
```diff
@@ -12,14 +12,11 @@ class Flowcell < ApplicationRecord
...
-    def self.includes_args(except = nil)
-      args = []
-      args << { library: Ont::Library.includes_args(:flowcell) } unless except == :library
-      args << { run: Ont::Run.includes_args(:flowcells) } unless except == :run
-
-      args
+    # Make table read only. We don't want anything pushing to it.
+    def readonly?
+      true
     end
```
JamesGlover, line 20: I wonder if it would be safer to call `readonly!` in an after_initialize call? That would have the advantage that we also set the `@readonly` instance variable. Doesn't matter too much, as it should all be both temporary and precautionary,.

JamesGlover (approved):
A few suggestions, and a couple of questions, but nothing critical.

PR #755, opened 2022-06-09
Issue752/dpl 396 as a team lead {james} i would like to create libraries without the qc data so we can print barcodes in the lab and freeze them

### LINKED ISSUE #752: DPL-396 - As a team lead (James) I would like to create libraries without the QC data so we can print barcodes in the lab and freeze them
This represents the traction-service work required for https://github.com/sanger/traction-ui/issues/783

Original issue:
**User story**
As a team lead (James) I would like to create libraries without the QC data so we can print barcodes in the lab and freeze them

Ideally we need to be able to print the TRAC-X-XXX barcodes (available on the libraries tab) for libraries before we freeze the samples in the lab
- But we cannot create a library without all QC data entered in ‘Create Library’ dialog
- This QC data can be delayed and not received until after the library needs to have been frozen (insert size) – preventing barcodes being printed in time
- We would like to be able to create a library without QC data

James

**Acceptance criteria**
- [ ] When creating a library I should not be prevented from doing so because the insert size has not been entered
- [ ] Prevention from using any library in a run unless the insert size has been added

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
Is all of the information still required for library creation? i.e. volume, concentration and template prep kit box barcode

JamesGlover:
Closes #752

* Removes validation of presence of volume, concentration, insert size on library creation
* Adds run_creation validation scope which enforces this instead
* Refines validation on volume, concentration, insert size
* Validates libraries are valid for run creation on run creation
* Adds field describing suitability for run creation to API

### DIFF
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
+require './app/resources/v1/shared/run_suitability'
+
 module V1
   module Pacbio
     # LibraryResource
     class LibraryResource < JSONAPI::Resource
+      include RunSuitability
+
       model_name 'Pacbio::Library'
 
       attributes :state, :volume, :concentration, :template_prep_kit_box_barcode,
--- a/spec/models/pacbio/library_factory_spec.rb
+++ b/spec/models/pacbio/library_factory_spec.rb
@@ -16,45 +16,46 @@ RSpec.describe Pacbio::LibraryFactory, type: :model, pacbio: true do
 
   context 'LibraryFactory' do
     describe '#initialize' do
-      before do
-        @factory = described_class.new(libraries_attributes)
-        @pacbio_library = @factory.library
-      end
+      subject(:factory) { described_class.new(libraries_attributes) }
+
+      let(:pacbio_library) { factory.library }
 
       it 'creates a Pacbio::Library object', aggregate_failures: true do
-        expect(@pacbio_library.volume).to be_present
-        expect(@pacbio_library.concentration).to be_present
-        expect(@pacbio_library.template_prep_kit_box_barcode).to be_present
-        expect(@pacbio_library.insert_size).to be_present
-        expect(@pacbio_library.id).to be_nil
-        expect(@pacbio_library.created_at).to be_nil
-        expect(@pacbio_library.updated_at).to be_nil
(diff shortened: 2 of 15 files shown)


app/resources/v1/pacbio/library_resource.rb
```diff
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
+require './app/resources/v1/shared/run_suitability'
```
stevieing, line 3: Why do we need to require this? Is there an issue with loading.

spec/models/pacbio/library_factory_spec.rb
```diff
@@ -114,20 +113,28 @@
       end
 
       context 'when save errors' do
+        subject(:factory) { described_class.new(attributes) }
+
         before do
-          @factory = described_class.new(attributes)
-          allow(@factory).to receive(:valid?).and_return true
+          # This test seems to be designed to catch edge cases, in which problems that
```
stevieing, line 119: I'm not sure what this means. Not clear why we need to stub out the valid method. Is it just a spurious test that can be removed?

stevieing, line 119: I'm not sure what this means. Not clear why we need to stub out the valid method. Is it just a spurious test that can be removed?

stevieing (approved):

app/resources/v1/pacbio/library_resource.rb
JamesGlover: There was, but looking at it now, I think its just because I haven't also name-spaced the module itself under shared.

spec/models/pacbio/library_factory_spec.rb
JamesGlover: Possibly... certainly as it stands its not hugely useful, but I wasn't really certain why it had been written. 

I think it is meant to be the belt to the braces of the validation (Ie. make sure nothing happend, even if we had failed to handle something explicitly), but I think there are better tests we could put in its stead, which is why I left the todo in there. Essentially it probably makes sense:

1) To check that nothing happens if either resource is invalid
2) To check that nothing happens if one of the saves fails for other reasons

I left it for now, as this was mostly something unrelated that rubocop picked up, and I wanted to keep this PR focused, hence popping in the issue alongside the todo.

PR #757, opened 2022-06-21
Issue709/dpl 277 ont create samples and requests {c m v 4}

JamesGlover:
Closes #

*
*
* ...


app/models/reception/request_factory.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
```
stevieing, line 1: extra comment

db/migrate/20220610094346_add_uniqueness_contrataint_to_library_type_name.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Rubocop want us to add a uniqueness constraint to the database. In practice
+# we're unlikely to reap much form the performance or data-integrity issues,
+# this will be a low volume rarely changing table, but there could be risks
+# of concurrency problems running tasks on deployment.
+class AddUniquenessContrataintToLibraryTypeName < ActiveRecord::Migration[7.0]
```
stevieing, line 7: Should this be AddUniquenessConstaintToLibraryTypeName. I know it is only semantics but probably worth changing now?

app/models/ont/library.rb
```diff
@@ -41,5 +39,11 @@ def assigned_to_flowcell
     def readonly?
```
stevieing, line 39: Does this need to be removed?

app/models/reception.rb
```diff
@@ -0,0 +1,26 @@
...
+  # Anything going via traction-ui (ie. everything initially) should be prefixed
+  # traction-ui. For example traction-ui.sequencescape or traction-ui.samples-extraction
+  # Validation here is a bit strict to maintain consistency. Especially if we end up with
+  # external collaberators
+  validates :source, presence: true, format: /\A[a-z0-9\-.]+\z/
+
+  delegate :request_attributes=, :construct_resources!, to: :resource_factory
+
+  # We flatten the keys here as they map back directly to the correpsonding
+  # attributes in Reception.We're mere'l using the ractory to encapsulate the behaviour
```
stevieing, line 18: should this be "We're merely using the factory"

app/models/reception/resource_factory.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
```
stevieing, line 1: It might be worth adding some more documentation here about why you are introducing caches. I think you mentioned it the other day so I was aware but it might be slightly opaque.

db/migrate/20220609141826_create_data_types_table.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Will be associated with ONT requests, but adding a pipline key to allow
```
stevieing, line 3: In this case it seems logical. We are coming from a place of experience so a bit of forward thinking is ok.

db/migrate/20220616114645_add_missing_indicies.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# We avoid creating duplicate resources, so lets an in index to help the reception
```
stevieing, line 3: "Let's add an index"... but makes sense and fixes another bug?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
```
stevieing, line 10: When I run the migrations I gewt the following warnings:

Loading application environment...
Loading code in search of Active Record models...
Generating Entity-Relationship Diagram for 39 models...
Warning: Ignoring invalid model ActiveStorage::Blob (table active_storage_blobs does not exist)
Warning: Ignoring invalid model ActiveStorage::VariantRecord (table active_storage_variant_records does not exist)
Warning: Ignoring invalid model ActiveStorage::Attachment (table active_storage_attachments does not exist)
Warning: Ignoring invalid association :taggable on TagTaggable (polymorphic interface Taggable does not exist)
Warning: Ignoring invalid association :container_materials on Ont::Library (Could not find the association :requests in model Ont::Library)
Warning: Ignoring invalid association :source_wells on Ont::Library (undefined method `klass' for nil:NilClass)
Done! Saved diagram to erd.jpg.

app/models/reception.rb
sdjmchattie:
```suggestion
  # attributes in Reception. We're merely using the factory to encapsulate the behaviour.
```

app/models/sample.rb
```diff
@@ -6,8 +6,9 @@ class Sample < ApplicationRecord
 
   attr_readonly :name
 
-  validates :name, :external_id, :species, presence: true
-  validates :name, uniqueness: { case_sensitive: false }
+  validates :species, presence: true
+  validates :external_id, presence: true, uuid: true
+  validates :name, uniqueness: { case_sensitive: false }, presence: true
```
sdjmchattie, line 11: Much clearer 👍

config/feature_flags.yml
```diff
@@ -0,0 +1,6 @@
+# Add feature flags to this file, example:
+# dpl202_enable_angry_cat_feature: When enabled this will add an angry cat to all pages that will try to catch the mouse pointer
```
sdjmchattie, line 2: Did you come up with this or is this in the example documentation from Flipper?  Sounds hilarious if someone implemented this.

db/schema.rb
JamesGlover: Warning: Ignoring invalid model ActiveStorage::Blob (table active_storage_blobs does not exist)
Warning: Ignoring invalid model ActiveStorage::VariantRecord (table active_storage_variant_records does not exist)
Warning: Ignoring invalid model ActiveStorage::Attachment (table active_storage_attachments does not exist)

Predate these changes, the others relate to the remains of the Ont libraries, can have a look to see if I can remove the bits that re causing it to get confused though.

config/feature_flags.yml
JamesGlover: My example. Wanted something that was obviously for demonstration purposes to avoid someone accidentally stripping it out thinking it was a delivered feature.

JamesGlover:
@sdjmchattie Yep, the raises are to validate that the code is indeed unused once the flags have been flipped. Intent is to help improve confidence in their removal, but to allow quick rollback if it turns out they are used elsewhere.

stevieing:
> I am a little uncomfortable with all the raises for when the flag is on, but I assume you've got them in the right places. I also assume it's just to check that the wrong behaviour isn't running when the flag is on. Normally you try to avoid this mixed behaviour throughout the whole system when using flags by making them smaller scope or making them able to direct all alternative behaviour through common areas of the code.

I get what you are saying but this made sense to me. There is a feature flag for the unified reception and other feature flags for turning off each of the pipeline receptions.

stevieing (approved):

stevieing (approved):

PR #758, opened 2022-06-21
Ruby 3

JamesGlover:
Closes #

*
*
* ...


JamesGlover:
See https://github.com/sanger/traction-service/pull/757 first

app/controllers/v1/pacbio/plates_controller.rb
```diff
@@ -21,8 +21,8 @@ def create
           render json: body, status: :created
         else
           # JSON-API formatted errors.
-          errors = @plate_creator.errors.map { |title, detail| { title: title, detail: detail } }
-          render json: { errors: errors },
```
JamesGlover, line 25: A lot of this, this is the new ruby hash feature, which lets you specify just a symbolic key for cases where the value is held in a variable of the same name.

Think I prefer the JS version of this syntax, but I guess this does have parallels with keyword arguments.

app/services/well_sorter_service.rb
```diff
@@ -3,9 +3,9 @@
 # Derived from https://github.com/sanger/code_snippets/blob/main/ruby/well_helpers.rb
 # Provides tools for dealing with well ranges
 module WellSorterService
-  COLUMNS_RANGE = { 96 => (1..12).freeze, 384 => (1..24).freeze }.freeze
+  COLUMNS_RANGE = { 96 => (1..12), 384 => (1..24) }.freeze
 
-  ROWS_RANGE = { 96 => ('A'..'H').freeze, 384 => ('A'..'P').freeze }.freeze
+  ROWS_RANGE = { 96 => ('A'..'H'), 384 => ('A'..'P') }.freeze
```
JamesGlover, line 8: Ranges (or possibly just range literals) are now frozen by default

stevieing (approved):

PR #788, opened 2022-07-15
Dpl 278 ont reception import a plate or tube from sequencescape or samples extraction with a library type

JamesGlover:
Closes #

* Add library type resource
* Add data type resource


app/controllers/v1/data_types_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module V1
+  # See the DataTypeResource for more information
```
stevieing, line 4: Might be worth providing a bit more explanation on data types. I can see what it is doing but not quite sure of it's significance.

lib/tasks/create_data_types.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
stevieing, line 1: Migth be worth adding an arbitrary test for the tasks just to ensure that we catch any code drift that might break the task. I have added them in my branch just a test to ensure that stuff is created. Also worth adding instructions in the readme.

lib/tasks/create_library_types.rake
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
stevieing, line 1: Again maybe worth adding a test. This came out of Seena setting up service as all of the tasks were broken. Also worth adding it to readme.

stevieing (changes requested):
All good. Mostly documentation and test things. Would it be worth adding the data tasks to bin/setup?
I haven't tested it locally but will do when I look at the ui.

lib/tasks/create_data_types.rake
stevieing: Also worth adding some feedback once the task has completed.

lib/tasks/create_library_types.rake
stevieing: Again also worth adding some feedback once the task has completed successfully.

stevieing (approved):

PR #792, opened 2022-07-29
dpl_277 - Cleanup disabled features and flags

JamesGlover:
**DO NOT MERGE UNTIL general reception is established**

This follows up dpl277 with the removal of the flagged features.

It also updates the pacbio rake task to use the reception factory, rather than mimicking the old mechanisms of reception.


BenTopping:
This pull request appears to be broken, I will regenerate one in its place

PR #802, opened 2022-08-18
DPL-439 QC Results

KatyTaylor:
Create QcAssayType and QcResult models
- haven't included the schema yet as have to sort out SQLite vs MySQL local changes

Closes #

*
*
* ...


lib/tasks/create_qc_assay_types.rake
```diff
@@ -0,0 +1,32 @@
...
+      { key: 'yield_post_0.45x_clean_up', label: 'Yield post 0.45X clean up', units: '' },
+      { key: 'qubit_post_45x_clean_up', label: 'Qubit post 0.45X clean up', units: 'ng/μl' },
+      { key: 'mode_length_dna_sheared_pacbio', label: 'Mode length of DNA sheared for PacBio', units: 'bp' },
+      { key: 'gqn_threshold_10kb', label: 'GQN with threshold at 10Kb', units: '' },
+      { key: 'qubit_post_shear_spri_45.4μl', label: 'Qubit (post shear:SPRI ng/μl in 45.4μl)', units: 'ng/μl' },
+      { key: '260_280_ratio_post_shear_spri', label: '260/280 ratio post shear:SPRI', units: '' },
+      { key: '260_230_ratio_post_shear_spri', label: '260/230 ratio post shear:SPRI', units: '' },
+      { key: 'pre_library_yield', label: 'Pre-library yield', units: '' },
+      { key: 'percent_dna_recovery_post_spri_pre_shear', label: '% DNA recovery (into shear - after SPRI)', units: '%' },
+      { key: 'shear_spri_result', label: 'Shear/SPRI result (pass/fail/hold)', units: '' }
```
emrojo, line 26: I see there are a lot of '' empty strings in units, and I understand the reason is because these assay types don't have a unit because they are ratios, booleans, etc... The other option would have been using NULL (nil). 

I think it will be usfule to add the description of the valid values in the model class, so we can understand what empty string means.

emrojo (approved):
All looks good, only one suggestion of comment

sdjmchattie (approved):

lib/tasks/create_qc_assay_types.rake
stevieing: I am going to change it to nil. Makes more sense. Also need to change the migrations to ensure there is a database constraint on the required fields.

PR #844, opened 2022-10-28
DPL-478 QC results upload

### LINKED ISSUE #815: DPL-478 - As a Developer I need an API endpoint to allow users to upload a QC results spreadsheet
**User story**
As a Developer I need an API endpoint to allow users to upload a qc results spreadsheet

Steve

This can be done within PSD. UAT would need to ensure that it still works with the front end

**Acceptance criteria**
- [x] Addition of status column to the qc results table. Decision on the QC outcome (free text field)
- [x] Addition of decision made by column to the qc results table to allow reporting of who made the status decision e.g Long Read or ToL so that decisions can be made independently by various stakeholders
- [x] An agreed csv format
- [x] The creation of a create only JSON API compliant endpoint which will accept a csv
- [x] A model which will be a record of the csvs uploaded. Do we need to save the raw data?
- [x] An associated factory which will handle the csv data and transform it into qc results
- [x] Validation to ensure that the qc results are correct
- [x] JSON API resource which will provide feedback on success or provide errors
- [ ] An agreed message format to send the qc results to the warehouse
- [ ] A handler to send the message to the warehouse
- [ ] modification of qc type column so that it can recognise different label (OPTIONAL - to decide)

**Dependencies**

- sanger/unified_warehouse#397
- sanger/traction-ui#885

**References**
- sanger/traction-ui#855

**Additional context**

- The above story is the originator but has been split out
- We will need the agreed csv format before progressing
- It is possible to split this story into sub stories e.g. creating the message structure

harrietc52:
Closes #815 
Coupled with https://github.com/sanger/traction-ui/pull/979


stevieing (commented):
Just thought I would have a quick look to see how you are getting on. I haven't gone into too much detail but it looks like you are on the right track.

After various conversations over the past couple of weeks it is looking more and more likely that we will need to reuse this for other qc results as an interim solution. Maybe for the SPK3 chemistry and sample prep.

On first glance maybe that would be done by delegating from the upload factory to a specific model based on a flag. I am not suggesting we refactor this at this point but maybe we can discuss this when we do the code review? Just to make sure it is flexible enought to do that.

lib/tasks/create_qc_assay_types.rake
```diff
@@ -1,29 +1,26 @@
...
+      { key: 'qubit_concentration_ngul', label: 'Qubit DNA Quant (ng/ul) [ESP1]', used_by: 0 },
+      { key: 'volume_si', label: 'DNA vol (ul)', used_by: 0 },
+      { key: 'yield', label: 'DNA total ng [ESP1]', used_by: 0 },
+      { key: '_260_230_ratio', label: 'ND 260/230 [ESP1]', used_by: 0 },
+      { key: '_260_280_ratio', label: 'ND 260/280 [ESP1]', used_by: 0 },
+      { key: 'nanodrop_concentration_ngul', label: 'ND Quant (ng/ul) [ESP1]', used_by: 0 },
+      { key: 'average_fragment_size', label: 'Femto Frag Size [ESP1]', used_by: 0 },
+      { key: 'gqn_dnaex', label: 'GQN >30000 [ESP1]', used_by: 0 },
+      { key: 'results_pdf', label: 'Femto pdf [ESP1]', used_by: 0 },
+      { key: 'some_future_key', label: 'Some Future Label', used_by: 1 }
```
harrietc52, line 23: @stevieing I've added another `used_by` option to test for reuse (following the steps at top of this file). The factory only persisted the relevant QC entities. Think that should cover the reuse use case?

app/models/qc_assay_type.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class QcAssayType < ApplicationRecord
+  enum used_by: { extraction: 0, some_future_group: 1 }
```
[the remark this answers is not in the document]
harrietc52: Sounds good, can always make it a table later if needed

app/validators/qc_results_upload_validator.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
```
stevieing, line 1: Nice :)

stevieing (approved):
Really good. I notice there are a few comments about doing further refactoring. You are probably right there is always more you could do but I think this is more than enough for now.

PR #848, opened 2022-11-02
DPL-279 - Adding ONT pool and libraries

BenTopping:
Initial work on DPL-279, adding ont pool and library models and resources.


app/models/ont/pool.rb
```diff
@@ -0,0 +1,55 @@
...
+    # after that is defined
+    include PlateSourcedLibrary
+    validates :volume, :concentration,
+              :insert_size, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
+    validates :libraries, presence: true
+    validates_with TagValidator
+
+    before_save { self.final_library_amount = calculate_final_library_amount }
+
+    def library_attributes=(library_options)
```
stevieing, line 21: There are a few methods here which are exactly the same in Pacbio so is there an opportunity for sharing or is it too early?

app/resources/v1/ont/library_resource.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
```
stevieing, line 2: Again wondering whether there is an opportunity to share any behaviour? I am not sure especially as it is volatile.

app/resources/v1/ont/pool_resource.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
```
stevieing, line 2: Same as with library resource.

spec/models/ont/flowcell_spec.rb
```diff
@@ -3,52 +3,54 @@
...
-    it 'does not destroy the associated run' do
-      run = flowcell.run
-      initial_flowcell_count = run.flowcells.count
-      flowcell.destroy
-      expect(run).not_to be_destroyed
-      expect(run.flowcells.count).to be(initial_flowcell_count - 1)
-    end
-  end
-end
+# RSpec.describe Ont::Flowcell, type: :model, ont: true do
```
stevieing, line 6: It might be better to skip these tests rather than comment them out? Unless there is a reason we can't do that?

spec/models/ont/run_spec.rb
```diff
@@ -3,88 +3,90 @@
 require 'rails_helper'
 require './spec/support/read_only'
```
stevieing, line 5: Same here. Would it be worth skipping these as well?

spec/pipelines/ont/message_spec.rb
```diff
@@ -3,131 +3,132 @@
 require 'rails_helper'
 require './spec/support/read_only'
```
stevieing, line 4: Again I would maybe think about skipping this.

stevieing (changes requested):
Pretty good. Just a few things some which I am not sure about.

Also as we discussed maybe worth adding some pending tests for the pagination.

app/resources/v1/ont/library_resource.rb
BenTopping: There probably is an opportunity to abstract the shared behaviour but as you said the models may change in the coming weeks. Can have a look back and refactor after the work is solid.

app/models/ont/pool.rb
BenTopping: Yes I found something where you could assign an 'attribute' with a default value but I couldnt figure out a way to call any class methods like calculate_final_library_amount from the default assignment. Think it was an issue with the lambda scope but i couldnt figure it out.

app/models/ont/pool.rb
```diff
@@ -0,0 +1,55 @@
...
+    has_many :requests, through: :libraries
+    # This is dependent on the requests association, so needs to be included
+    # after that is defined
+    include PlateSourcedLibrary
+    validates :volume, :concentration,
+              :insert_size, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
+    validates :libraries, presence: true
+    validates_with TagValidator
+
+    before_save { self.final_library_amount = calculate_final_library_amount }
```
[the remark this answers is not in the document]
BenTopping: Yes I found something where you could assign an 'attribute' with a default value but I couldnt figure out a way to call any class methods like calculate_final_library_amount from the default assignment. Think it was an issue with the lambda scope but i couldnt figure it out.

app/models/ont/pool.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def missing_library(id)
+      raise ActiveRecord::RecordNotFound, "Ont request #{id} is not part of the pool"
+    end
+
+    def indexed_libraries
+      @indexed_libraries ||= libraries.index_by { |lib| lib.id.to_s }
+    end
+
+    def calculate_final_library_amount
```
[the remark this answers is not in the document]
BenTopping: Added some comments and constants 👍

app/models/ont/pool.rb
```diff
@@ -0,0 +1,55 @@
...
+      id = attributes['id'].to_s
+      indexed_libraries.fetch(id) { missing_library(id) }
+                       .tap { |l| l.update(attributes) }
+    end
+
+    def missing_library(id)
+      raise ActiveRecord::RecordNotFound, "Ont request #{id} is not part of the pool"
+    end
+
+    def indexed_libraries
```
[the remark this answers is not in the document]
BenTopping: Im not 100% sure either so dont want to comment something that is wrong but I think it is a helper method used when updating libraries within a pool

app/models/ont/pool.rb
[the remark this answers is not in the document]
stevieing: OK. Maybe worth adding a comment for now to that effect and we may have a lightbulb moment and find a fix?

app/resources/v1/ont/library_resource.rb
stevieing: Agreed.

stevieing (approved):

app/resources/v1/ont/pool_resource.rb
```diff
@@ -0,0 +1,54 @@
...
+      def created_at
+        @model.created_at.to_fs(:us)
+      end
+
+      def updated_at
+        @model.updated_at.to_fs(:us)
+      end
+
+      # def publish_messages
```
stevieing, line 49: Is this spurious? I presume we are not processing messages for pools yet!

db/migrate/20221102100727_drop_old_ont_tables.rb
```diff
@@ -0,0 +1,11 @@
+class DropOldOntTables < ActiveRecord::Migration[7.0]
```
stevieing, line 1: Do we need to drop all of these? ont_runs and ont_flowcells are going to be used? Maybe there is a specific reason?

stevieing (changes requested):
Pretty much there. A couple of minor things as I reviewed it previously.
Needs linting and fixing merge conflicts.

db/migrate/20221102100727_drop_old_ont_tables.rb
BenTopping: I dropped them to nuke the old data and because I thought the run data would most likely change from the current tables. Happy to keep them though, not sure whats best

app/resources/v1/ont/pool_resource.rb
BenTopping: Yes think this was copied over and commented out, will most likely be added in future which is why it was kept

db/migrate/20221102100727_drop_old_ont_tables.rb
stevieing: OK. That's fine. Doesn't matter either way.

stevieing (approved):

PR #868, opened 2022-12-14
DPL-281 ONT flowcells, instruments, and runs.

yoldas:
Initial work on DPL-281. Added flowcells, instruments, and runs.


app/models/ont/flowcell.rb
```diff
@@ -3,21 +3,26 @@
...
     include Uuidable
 
+    # Run has multiple of these flowcells up to the maximum number for the instrument.
     belongs_to :run, foreign_key: :ont_run_id, inverse_of: :flowcells
-    belongs_to :library, foreign_key: :ont_library_id, inverse_of: :flowcell
-    delegate :requests, to: :library
 
+    # We have changed the model to use pool rather than library.
+    # We assume one-to-one relationship with pool at the moment.
+    # belongs_to :library, foreign_key: :ont_library_id, inverse_of: :flowcell
```
BenTopping, line 13: Remove extra comments in this file that aren't needed.

app/models/ont/run.rb
```diff
@@ -3,12 +3,69 @@
...
+        if pool_set.include? flowcell.ont_pool_id
+          errors.add(:flowcells, "pool #{flowcell.ont_pool_id} is duplicated in the same run")
+        else
+          pool_set.add(flowcell.ont_pool_id)
+        end
+      end
+    end
+
+    def check_max_number_of_flowcells
+      return if instrument.blank?
```
BenTopping, line 59: can merge these guard clauses into one something like return if instrument.blank? || flowcells.length <= intrument.max_number_of_flowcells

BenTopping: also if you validate instrument's presence you wouldnt need the instrument.blanks guard

app/models/ont/run.rb
```diff
@@ -3,12 +3,69 @@
...
+
+      return if flowcells.length <= instrument.max_number_of_flowcells
+
+      errors.add(:flowcells, 'must be less than instrument max number')
+    end
+
+    # Generate the experiment_name using the id of the run.
+    after_create :generate_experiment_name
```
BenTopping, line 67: instead of using after_create you might be able to use something like `attribute :experiment_name , default: -> { "#{NAME_PREFIX}#{id}" }` . If not i think using `before_save` is generally safer than after_create

spec/factories/ont/instruments.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :ont_instrument, class: 'Ont::Instrument' do
+    sequence(:name) { |n| "O#{n}" }
+    instrument_type { 'GridION' }
+    max_number_of_flowcells { 5 }
+
+    factory :ont_gridion do
```
BenTopping, line 9: Is this extra factory needed if you change the base factory to gridion?

spec/models/ont/instrument_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
BenTopping, line 1: maybe worth testing the max flowcells works by creating an instrument and trying to assign more flowcells than allowed? Not sure if thats tested elsewhere

spec/factories/ont/instruments.rb
yoldas: ont_instrument is a generic one. We could remove 'GridION' from that but instrument_type uses enum values.
ont_gridion is more specific and definitively with 5 max number of flowcells.
We are using both in tests.

BenTopping (approved):
Looking good 👍 Couple comments but i dont think any prevent merging if you disagree with them

app/models/ont/run.rb
BenTopping: Ah yes, makes sense. Can experiment_name just be a method attribute then? If its never renamed do we need to store it in the database? We could just have a method that returns `"#{NAME_PREFIX}#{id}"`

app/models/ont/flowcell.rb
yoldas: I have removed the extra comments.

app/models/ont/run.rb
yoldas: Instrument presence is validated implicitly by belongs_to.
Without the instrument we cannot use those some validators because of instrument.max_number_of_flowcells (attribute access on nil). If ont_instrument_id is not given, the client would get 500 error about attribute access on nil instead of missing instrument id.

The guard is not part of the validation. It checks whether that validation can be done.

app/models/ont/run.rb
yoldas: One of the Acceptance criteria requires persistence of the name.

yoldas:
> @yoldas Could you also update lib/tasks/ont_data rake task to also generate dummy runs. This will be useful for people trying to setup some test data

I have added 6 runs and their flowcells to the task based on the pools created earlier.

BenTopping (approved):
Looks good.

PR #883, opened 2023-01-05
715 dpl 285 ont sending sequencing run data to the warehouse

stevieing:
NB This includes the code from DPL-285. Probably not worth merging until that is completed so we can merge and resolve any conflicts.


config/pipelines.yml
```diff
@@ -481,28 +484,43 @@ default: &default
             instrument_slot:
               type: :model
               value: position
+            flowcell_id:
+              type: :model
+              value: flowcell_id
             samples:
               type: :array
-              value: requests
+              value: libraries
```
BenTopping, line 492: Are these libraries through pools? Just getting my head around the naming conventions and the actual data

stevieing: Yes. Samples is what gets sent to the warehouse in the message. We can't use pool as that does not contain the individual request information. And also yes the libraries are through pools.

PR #884, opened 2023-01-10
DPL-478 Qc Upload CSV fixes

harrietc52:


app/models/qc_results_upload_factory.rb
```diff
@@ -41,8 +41,10 @@ def create_entities!
 
   # @param [Object] CSV row e.g. { col_header_1: row_1_col_1 }
   def create_data(row_object)
-    # 1. Always create Long Read QC Decision
-    lr_qc_decison = create_qc_decision!(row_object[LR_DECISION_FIELD], :long_read)
+    # 1. If required, create Long Read QC Decision
+    if row_object[LR_DECISION_FIELD]
```
stevieing, line 45: There is a bit of duplication here. Now we are processing both decision types in the same way they could maybe be combined?

app/models/qc_results_upload_factory.rb
```diff
@@ -144,13 +149,16 @@ def qc_decision
...
-      assay_type ? assay_type.key : header
+      assay_type ? assay_type.key : header.strip
     end
 
-    csv = CSV.new(csv_string_without_groups, headers: true, header_converters: header_converter,
-                                             converters: :all)
+    csv = CSV.new(csv_string_without_groups, headers: true, header_converters: header_converter)
 
-    csv.to_a.map(&:to_hash)
+    arr_of_hashes = csv.to_a.map(&:to_hash)
```
stevieing, line 157: I think a bit of documentation would be good just to explain why we are excluding these lines?

db/migrate/20230109154242_modify_csv_data_in_qc_results_upload.rb
```diff
@@ -0,0 +1,5 @@
+class ModifyCsvDataInQcResultsUpload < ActiveRecord::Migration[7.0]
+  def change
```
stevieing, line 2: Longtext - new one on me! I guess that is much better to manage than a BLOB.

stevieing (approved):
I have added a couple of things which would be useful to change but they are not crucial. Didn't want to hold you up any more. Maybe worth at least adding comments for the future?

app/models/qc_results_upload_factory.rb
harrietc52: I've pulled it out it to its own method, but agree some refactoring could be done. Had a quick try but it wasn't coming to me so will leave it for now and get it into UAT, but will keep pondering on it

stevieing: OK. I had an idea of how to do it but it can wait.

harrietc52:
```
        let errors = e.split(',')
        let count = errors.length
        let uniqueErrors = [...new Set(errors)]

        let msg = `${count} rows have error: ${uniqueErrors}`

        // 195 rows have error: Missing data: LR EXTRACTION DECISION [ESP1] csv_data - Missing data: LR EXTRACTION DECISION [ESP1], Missing data: LR EXTRACTION DECISION [ESP1] csv_data - Missing data: LR EXTRACTION DECISION [ESP1]
        this.showAlert(msg, 'danger')
```

PR #890, opened 2023-01-13
DPL-281 Updates

yoldas:
Closes #

- [x] Added tests for error responses for ONT Run creation.
- [x] Added validation for min number of flowcells on ont run [moved in refactoring]
- [x] Test for Messages not published on failure
- [x] Refactoring to use built-in validations [where possible]
- [x] Conditional validation of flowcell positions based on presence of run and instrument.
- [x] Uniqueness validation of flowcell position (unique in run)
- [x] Uniqueness validation of flowcell pool [Removed the uniqueness validation as they can be duplicated]
- [x] Add pagination to runs resource
- [x] Add validations for flowcell_id
- [x] Create migration for unique together (run, pool) index for pool [Removed the constraint now as pools can be used in multiple flowcells]
- [x] Create migration for unique index for flowcell_id [converted to unique-together later]
- [x] Update ont_flowcell factory to generate unique flowcell_id values [converted to unique-together later]
- [x] Update ont_data rake task for unique flowcell_id values
- [x] Validations on the run for flowcells
- [x] Ensure no 500 errors for known cases
- [x] Fix position range validation
- [x] Date format for ONT run resource created_at same as Pacbio run resource.
- [x] Fix the exception in flow cell_attributes assignment [primarily affected update requests, caused by using CollectionProxy's flowcells= method]. 

After demo:
- [x] - Service: Pagination
- [x] - Service: When there is an error on run create/ update, provide information about the position of the invalid flowcell. (e.g. Currently says pool is not known, instead say pool at position X is not known.)
- [x] - Service: Want to allow same pool on more than one flowcell in the same run
- [x] - Service: Make Pool Barcode and Flowcell case insensitive on run create/ update 

[Service does not accept a "pool barcode" parameter; it accepts ont_pool_id; pool barcode is actually the barcode of the Tube of the Pool, which is used by the UI]
[Service can strip and uppercase Flowcell_id before validating and storing; that will be the easiest solution]

Additional:
- [x] Runs resource default sort order is by created_at descending
- [x] Remove flowcell_id index and add (run,flowcell_id) unique together index, change validations, etc.
- [x] Case-insensitive barcode filters
- [x] Locale strings for error messages
- [x] Flowcell_id position display 
- [x] Delegate max_number_of_flowcells to run
- [x] Strip and upcase flowcell_id using an attribute writer
- [x] Simplified library_attributes writer
- [x] Moved position names to Ont::Instrument
- [ ] Create a custom type for position **[We are trying to abstract the position to pass the position name to the client in a way without breaking existing code. ]**

Questions:
 Do we need to include updated_at field in runs resource? [No]


spec/requests/v1/ont/runs_spec.rb
```diff
@@ -60,6 +60,348 @@ def json
         expect(fc2.pool.id).to eq(attr2[:ont_pool_id])
       end
     end
+
+    context 'on failure' do
```
stevieing, line 64: Ah. Because you have checked each failure in the model you only need to test that the response is not successful and it has some errors. So you only need one failure test. Also as I have added message publishing which I have just merged into develop you will also need to check that a message is not published.

stevieing (changes requested):
You probably only need one failure test in the requests as it is thoroughly tested elsewhere. Sorry should have mentioned that!

spec/requests/v1/ont/runs_spec.rb
yoldas: I have removed the redundant tests, merged develop into the branch and added Messages-not-published tests to error cases.

sdjmchattie (approved):
I would wait for @stevieing review as he knows this system better than me, but I don't see anything that I consider a concern.

app/models/ont/flowcell.rb
```diff
@@ -35,32 +35,55 @@ class Flowcell < ApplicationRecord
...
+        PromethION: {
+          1 => '1A', 2 => '1B', 3 => '1C', 4 => '1D',
+          5 => '1E', 6 => '1F', 7 => '1G', 8 => '1H',
+          9 => '2A', 10 => '2B', 11 => '2C', 12 => '2D',
+          13 => '2E', 14 => '2F', 15 => '2G', 16 => '2H',
+          17 => '3A', 18 => '3B', 19 => '3C', 20 => '3D',
+          21 => '3E', 22 => '3F', 23 => '3G', 24 => '3H'
+        }
+      }
+    end
```
sdjmchattie, line 87: No problem with doing it this way for this number of locations, though I would have probably done this something like because it makes it less likely to contain a typo in this manually generated mapping:

```
device_ids = (1..3).flat_map do |i|
  ('A'..'H').flat_map do |j|
    "#{i}#{j}"
  end
end

@device_id_map ||= Hash[ device_ids.each_with_index.map { |v, i| [ i + 1, v ] } ]
```

app/models/ont/flowcell.rb
```diff
@@ -8,15 +8,82 @@ class Flowcell < ApplicationRecord
...
+      end
+    }
+    validates :position, uniqueness: {
+      scope: :run,
+      message: lambda do |_object, data|
+        "position #{data[:value]} is duplicated in the same run"
+      end
+    }
+
+    # pool validations
```
stevieing, line 35: Not entirely sure what is going on here? What is object.device_id?

app/models/ont/flowcell.rb
stevieing: I concur

app/models/ont/run.rb
```diff
@@ -10,57 +10,61 @@ class Run < ApplicationRecord
...
-        if flowcell.position < 1 || flowcell.position > instrument.max_number_of_flowcells
-          errors.add(:flowcells, "position #{flowcell.position} is out of range for the instrument")
-        elsif position_set.include? flowcell.position
-          errors.add(:flowcells, "position #{flowcell.position} is duplicated in the same run")
-        else
-          position_set.add(flowcell.position)
-        end
+    # run flowcells validations
+    validates :flowcells, presence: true
+    validates :flowcells, length: {
```
stevieing, line 25: Again I am not entirely sure why you need a block for this?. The message is fixed so can't you just have the message as a string?

lib/tasks/ont_data.rake
```diff
@@ -121,30 +123,33 @@ namespace :ont_data do
         break if safe_peek(pool_enum).blank?
 
         position = position_cycle.next
-        flowcell_id = format('F%05d', position)
-        Ont::Flowcell.new(flowcell_id:, position:, run:, pool: pool_enum.next)
+        pool = pool_enum.next
+        flowcell_id = format('F%05d', flowcell_id_enum.next)
```
harrietc52, line 127: What does this line do? I have some validation in the front end to ensure the Flowcell ID has: 3 letters followed by at least 3 numbers. 

I think this is what Alex specified, that a flowcell looks something like `ABC123`

app/models/ont/run.rb
```diff
@@ -77,29 +81,60 @@ def cancel
...
           update_flowcell(attributes)
         else
-          Ont::Flowcell.new(attributes)
+          create_flowcell(attributes)
         end
       end
     end
 
+    # Returns error messages added so far
+    def errors_messages
```
stevieing, line 101: Can you use errors.full_messages?

stevieing (changes requested):
It might be worth adding some documentation to the validations and methods that you have added as I am not entirely sure what they are doin.

spec/models/ont/flowcell_spec.rb
```diff
@@ -62,4 +72,53 @@
...
+    it 'returns coordinates for PromethION' do
+      run = create(:ont_promethion_run, flowcell_count: 24)
+      first = run.flowcells.find_by(position: 1)
+      last = run.flowcells.find_by(position: 24)
+
+      expect(first.position).to eq(1)
+      expect(first.device_id).to eq('1A')
+
+      expect(last.position).to eq(24)
+      expect(last.device_id).to eq('3H')
```
harrietc52, line 86: Thank you 👍 maybe it could be called `coordinate` instead of `device_id`? the `_id` might be confusing

lib/tasks/ont_data.rake
harrietc52: Also noticed the flowcell validation `F%05d` is specified in a few places, maybe that could get pulled out somewhere?

spec/requests/v1/ont/pools_spec.rb
```diff
@@ -392,4 +394,29 @@
...
+      barcodes = pools.map { |pool| pool.tube.barcode }
+      downcased = barcodes.map(&:downcase)
+      get "#{v1_ont_pools_path}?filter[barcode]=#{downcased.join(',')}", headers: json_api_headers
+
+      expect(response).to have_http_status(:success)
+      expect(json['data'].length).to eq(barcodes.length)
+      expect(json['data'][0]['attributes']['tube_barcode']).to eq barcodes[0]
+      expect(json['data'][1]['attributes']['tube_barcode']).to eq barcodes[1]
+    end
+  end
```
harrietc52, line 421: Nice 👍

spec/requests/v1/ont/runs_spec.rb
```diff
@@ -148,28 +187,185 @@ def json
...
+
+        let(:body) do
+          {
+            data: {
+              type: 'runs',
+              id: run.id,
+              attributes: {
+                ont_instrument_id: run.ont_instrument_id,
+                state: 'completed'
+              }
```
harrietc52, line 202: Do you want to test the run creation with any flowcell data too maybe. Would this flowcell data also be sent to the MLWH?

harrietc52 (commented):
Nice, really thorough in your validations and testing. Added a few comments. Only important comment is the Flowcell ID format I think

app/models/ont/flowcell.rb
yoldas: UI asked to give coordinates for PromethION flowcells, instead of numbers, in error messages. device_id either returns a coordinate such as 1A or a position.

app/models/ont/flowcell.rb
yoldas: I have used this piece of code to generate coordinates.

app/models/ont/run.rb
```diff
@@ -77,29 +81,59 @@ def cancel
...
+        raise ActiveRecord::RecordNotFound, "Ont flowcell #{id} does not exist"
+      else
+        raise ActiveRecord::RecordNotUnique, "Ont flowcell #{id} is duplicated"
+      end
     end
 
-    def missing_flowcell(id)
-      raise ActiveRecord::RecordNotFound, "Ont flowcell #{id} does not exist"
+    def create_flowcell(attributes)
+      new_flowcell = flowcells.build
```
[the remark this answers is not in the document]
yoldas: Nested objects are saved together when run is saved. I update or build them, they are validated together and saved together.

app/models/ont/flowcell.rb
andrewsparkes: fyi Coordinates will be row then column, so A1 not 1A

app/models/ont/flowcell.rb
```diff
@@ -8,15 +8,59 @@ class Flowcell < ApplicationRecord
...
+    validates :ont_pool_id, presence: {
+      message: lambda do |object, _data|
+        "pool at position #{object.position} is unknown"
```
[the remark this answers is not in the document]
yoldas: The first one checks ont_pool_id presence, the second checks if there is really a pool with that id.
The first message should have said "is required" or "is missing". However, when UI is relating tube_barcodes to pool ids, id comes out as nil. The interface shows a bad value but the error message for that value would be shown as "is required". Putting "is unknown" for that too looked better.

app/models/ont/flowcell.rb
yoldas: Thanks for the code. I have changed the occurrences of device_id to position_display.

app/models/ont/flowcell.rb
yoldas: I have replaced the occurrences of device_id with position_display now. 
I have change the places of letter and number for promethION.

app/models/ont/run.rb
yoldas: I have removed the lambda block for messages and added rubocop:disable Rails/I18nLocaleTexts .

app/models/ont/run.rb
yoldas: I have removed the redundant method and duplicate message checking.

app/models/ont/flowcell.rb
```diff
@@ -8,15 +8,59 @@ class Flowcell < ApplicationRecord
...
+    validates :flowcell_id, uniqueness: {
+      scope: :run,
+      message: lambda do |object, data|
+        "flowcell_id #{data[:value]} at position #{object.position} is duplicated in the same run"
+      end
+    }
+
+    # Return the max_number of flowcells for the instrument if run and instrument are present.
+    def max_number_of_flowcells
+      run.present? && run.instrument.present? && run.instrument.max_number_of_flowcells
```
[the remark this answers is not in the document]
yoldas: I have applied both suggestions.

app/models/ont/run.rb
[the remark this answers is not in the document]
yoldas: I have added a comment about this for flowcell_attributes=

 This method is called before validations, therefore it must properly
  set up all flowcells to be saved together with the run. It must first
  do the deletions and then "build" flowcells and "assign" attributes
  without saving.

lib/tasks/ont_data.rake
yoldas: I have defined Ont::Flowcell::FLOWCELL_ID_FORMAT = 'ABC%03d' and used it in other places. It is used in factories and ont_data:create task.

spec/models/ont/flowcell_spec.rb
yoldas: I have already renamed it to position_display before this suggestion. I think, singular and plural for coordinate is confusing. The method can also return a number (position) if there is no mapping for a device as well.

app/models/ont/flowcell.rb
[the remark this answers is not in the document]
yoldas: I have removed the separate ont_pool_id validation. Pool validation is enough.

app/models/ont/flowcell.rb
```diff
@@ -8,15 +8,82 @@ class Flowcell < ApplicationRecord
...
 
     delegate :requests, :libraries, to: :pool
 
-    # Validate the position to be a positive integer.
-    validates :position,
-              presence: true,
-              numericality: { greater_than_or_equal_to: 1 }
-    validates :flowcell_id, presence: true
+    # flowcell position validations
+    validates :position, presence: true
```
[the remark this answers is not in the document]
yoldas: I have added comments for the validations now.

stevieing (changes requested):
I have made quite a few comments. It might be worth having a catch up to go through it to explain what I mean.

app/models/ont/run.rb
```diff
@@ -76,30 +87,72 @@ def cancel
...
     end
 
     private
 
+    # Change attributes before using them
+    # Strip and upcase flowcell_ids to accept case-insensitive barcodes.
+    def transform_flowcell_attributes(flowcell_options)
+      flowcell_options.each do |attributes|
+        # because it is a frozen string
+        if attributes[:flowcell_id]
```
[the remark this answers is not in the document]
yoldas: This method is used by flowcell_attributes= method of run. The run is responsible for assignment of them as they are nested attributes of the run (autosave is enabled for flowcells; we create or update the run to persist flowcells together). In this particular case, it is sanitising the incoming data, converting flowcell_id (barcode-like) to capitals because we allowed case-insensitive input.

flowcell_attributes= is called before validations, we need to modify flowcells so the run and its flowcells are validated together. I will talk about it in the other comment as well.

app/models/ont/run.rb
```diff
@@ -76,30 +87,72 @@ def cancel
...
+          attributes[:flowcell_id] = attributes[:flowcell_id]&.strip&.upcase
+        end
+      end
+    end
+
+    # Assing attributes to an existing flowcell. The flowcell is found by the
+    # id given in the attributes. If there is no flowcell with that id, an
+    # exception is raised.
     def update_flowcell(attributes)
```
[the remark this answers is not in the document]
yoldas: We are receiving attributes and removing/changing/adding flowcells in flowcell_attributes= . I can move individual methods called by flowcell_attributes= to private block as they are not supposed to be used elsewhere.

app/models/ont/run.rb
```diff
@@ -76,30 +87,72 @@ def cancel
...
 
-    def indexed_flowcells
-      @indexed_flowcells ||= flowcells.index_by { |flowcell| flowcell.id.to_s }
+    # Create a new flowcell instance and assign attributes.
+    def create_flowcell(attributes)
+      new_flowcell = flowcells.build
+      assign_flowcell_attributes(new_flowcell, attributes)
+    end
+
+    def assign_flowcell_attributes(flowcell, attributes)
```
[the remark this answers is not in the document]
yoldas: The original implementation which was direct assignment to flowcells association inside flowcell_attributes= was failing because it did not consider deleting the existing ones first and did saving or updating of individual flowcells.

The new implementation deletes flowcells missing from the attributes, builds new ones, assign attributes. The run an and flowcells go to validation together and then saved. 

As for calling asking_flowcell_attributes for new and existing flowcells:
The method enables a uniform handling of flowcells and attributes
The method assigns only the attributes that flowcell can respond to (compared to mass-assignment or cherry-picking which attributes to assign.)

app/models/ont/run.rb
```diff
@@ -5,62 +5,73 @@ module Ont
...
+    validate :flowcells, :position_uniqueness
+
+    # flowcell_id uniqueness
+    # flowcell_id must be unique within the run. Previously this validation was
+    # on Flowcell as scoped uniqueness validation but it did not work as
+    # expected when creating run and flowcells together.
+    validate :flowcells, :flowcell_id_uniqueness
+
+    # Check if positions are duplicated in the run.
+    def position_uniqueness
```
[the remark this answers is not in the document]
yoldas: We are not trying to validate "each" of them here; we are trying to take them together to check duplicated position field in the same run.

app/models/ont/flowcell.rb
```diff
@@ -5,18 +5,106 @@ module Ont
...
+        )
+      }
+    }
+
+    # Return the max_number of flowcells for the instrument if run and instrument are present.
+    def max_number_of_flowcells
+      run&.max_number_of_flowcells
+    end
+
+    # Returns alternative adressing for position if available
```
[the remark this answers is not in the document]
yoldas: The map is only used by the flowcells to show "flowcell position" in a different format such as 1A or x1 in error messages. Although I don't see much use for that for the moment, we can move it to a separate module.

app/models/ont/flowcell.rb
```diff
@@ -5,18 +5,106 @@ module Ont
...
+    # pool validations
+    # pool must exist. The error message is given as "is unknown" because of
+    # how the UI shows the message. The UI gets a tube barcode and finds the
+    # corresponding pool. If it can't find the pool, it sends a nil
+    # ont_pool_id . When it shows the error message, the invalid tube barcode
+    # entered by the user is still shown in the page. If the error message was
+    # "is required" or "is missing", it would be confusing.
+    # We do not validate ont_pool_id separately as the presence of the pool
+    # covers it.
+    # We used generate_message to be able to pass options from this object.
```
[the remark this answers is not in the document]
yoldas: The UI needs to tell user at which position there is an error in different formats for different instrument runs. Standard errors were not enough. We need to put flowcell.position_display into the message.

In case or "errors.add" used in custom validation methods, it was relatively easy to pass the flowcell.position_display into the locale string, as it does the searching through the namespace in the locale file for the object.

However, I had to use generate_message to build the message when we are using built-in validations because I could not put flowcell.position_display into the scope of validation while handling the locale string. 

It also saves me from handling the locale string namespace handling. Direct uses of I18n was very messy. 

The errors.add method also use generate_message.

config/locales/en.yml
```diff
@@ -44,9 +44,24 @@ en:
         ont/library:
           attributes:
             name:
-              duplicated_in_plate: 'must be unique: a pool already exists for this plate'
+              duplicated_in_plate: "must be unique: a pool already exists for this plate"
```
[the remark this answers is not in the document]
yoldas: That locale string was already there. I saw an intention of quoting, but, in YAML, they need to be double-quoted or not quoted at all as far as I know. That is why it was a quick edit without much thinking.

stevieing (approved):
All good. We have already discussed future improvements and how we can reuse some of this across other pipelines. I am sure we will come back to it.

PR #921, opened 2023-02-14
Dpl642 config for warehouse

emrojo:
Closes #

*
*
* ...


app/resources/v1/reception_resource.rb
```diff
@@ -20,18 +20,31 @@ def request_attributes=(request_parameters)
...
+      publish_messages
     end
 
     def permitted_request_attributes
       [*::Pacbio.request_attributes, *::Ont.request_attributes, *::Saphyr.request_attributes].uniq
     end
+
+    def publish_messages
+      Messages.publish(
+        @model.requests.map(&:sample).flatten,
```
yoldas, line 41: Why do we need to flatten?

config/pipelines/reception.yml
```diff
@@ -0,0 +1,62 @@
...
+          common_name:
+            type: :model
+            value: species
+
+
+development: *default
+test: *default
+uat: *default
+training: *default
+production: *default
```
yoldas, line 62: What uses this configuration?
How are values like Time.current expanded?

emrojo: This configuration is the definition of the message payload going to the warehouse from Traction.
There is some code in ```app/messages/messages/message.rb``` that evaluates the expression of value before sending.

app/resources/v1/reception_resource.rb
emrojo: I think we don't because ```sample``` is not a list, so I'll remove it.

yoldas (approved):

PR #933, opened 2023-02-28
Revert "Revert "Dpl642 config for warehouse""

emrojo:
These changes were not tested in uat so they were reverted while waiting for uat


yoldas (approved):

Gemfile
```diff
@@ -32,6 +32,7 @@ end
 group :development, :test do
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
   gem 'pry-rails'
+  gem 'rspec-json_expectations'
```
stevieing, line 35: I am not sure that we should use this gem. It has not been updated for 4 years. This could cause technical debt in the future.

yoldas: This PR is to revert to merge. In the next PR, include_json matcher needs an alternative then; two tests need changing.

spec/support/json_matcher.rb
```diff
@@ -0,0 +1,149 @@
...
+# of the pipeline returned true, or if all the steps returned nil
+def run_checks_pipeline(object, compared_object, key)
+  pipelines = %w[
+    check_regexp check_subtree check_array check_value
+  ]
+  pipelines.each do |method_name|
+    value = send(method_name, object, compared_object, key)
+    return value unless value.nil?
+  end
+  true
```
yoldas, line 118: Each method in the pipelines array returns a boolean; true or false. I think it is never nil, which will cause early exit. I don't think we can reach the last line of the run_checks_pipeline method.

spec/support/spec/json_matcher_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  describe '#match_json' do
+    it 'can perform checks' do
+      expect(match_json('{"a": 1}', { 'a' => 1 })).to be_truthy
+      expect(match_json('{"a": 1}', { 'a' => 2 })).to be_falsy
+    end
+  end
+end
```
yoldas, line 58: It would be useful to have a couple of tests for hash and arrays; with different keys and lengths.

yoldas (commented):
I have added two comments.

spec/support/json_matcher.rb
emrojo: The first line of the checks have a 'return' without any arguments, which it will be nil when met the condition. nil means that the check does not apply.

spec/support/spec/json_matcher_spec.rb
emrojo: I'll add them

emrojo: Added

PR #944, opened 2023-03-08
Dpl685 avoid duplicates

emrojo:
Closes #

*
*
* ...

### DIFF
--- a/spec/validators/resource_factory_validator_spec.rb
+++ b/spec/validators/resource_factory_validator_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe 'ResourceFactoryValidator' do
           container: { type: 'tubes', barcode: 'NT1' }
         }]
       end
-    
+
       before do
         validator.validate(resource_factory)
       end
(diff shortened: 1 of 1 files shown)


app/models/well.rb
```diff
@@ -16,6 +16,11 @@ class Well < ApplicationRecord
   validates :position, presence: true
   validates :barcode, presence: true, on: :reception
 
+  scope :by_barcode, ->(*barcodes) { joins(:plate).where(plate: { barcode: barcodes }) }
```
stevieing, line 19: You can chain scopes but not sure if that is performant.

stevieing (changes requested):
Pretty good. Quite a simple solution. just one minor thing.

app/models/well.rb
emrojo: I think they would be the same, but I didnt see a use case where you would want to use individually a by_position() scope to extract all wells for the same position.

stevieing (approved):

PR #953, opened 2023-03-16
Cleanup old receptions

BenTopping:
James G:
"This follows up dpl277 with the removal of the flagged features.

It also updates the pacbio rake task to use the reception factory, rather than mimicking the old mechanisms of reception."


stevieing (approved):
Not a lot to say. Always good to see so much code removed.

PR #970, opened 2023-04-04
962 dpl 706 as developers we need to create a smrt link version 12 for revio so that users can create sample sheets

stevieing:
Closes #

*
*
* ...


enegoktan (approved):
Dropping and reseting the database now works as SmrtLink v11 is now set as the default 👍

PR #974, opened 2023-04-05
DPL-705 Prevent pacbio pulling back all records

BenTopping:

* Updated pacbio resources missing pagination. (plates, tubes and requests)
* Updated reception generator to handle iteration errors for tube creation

This PR is coupled with: https://github.com/sanger/traction-ui/pull/1110

### DIFF
--- a/lib/tasks/reception_generator.rb
+++ b/lib/tasks/reception_generator.rb
@@ -62,7 +62,7 @@ class ReceptionGenerator
     @number_of_tubes.times.map do
       barcode = @barcodes.next
       library_type = @library_types.next
-      data_type = @data_types.next
+      data_type = begin; @data_types.next; rescue StopIteration; nil; end
       {
         request: request(library_type, data_type),
         sample:,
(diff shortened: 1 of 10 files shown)


lib/tasks/reception_generator.rb
```diff
@@ -62,7 +62,7 @@ def tubes
     @number_of_tubes.times.map do
       barcode = @barcodes.next
       library_type = @library_types.next
-      data_type = @data_types.next
+      data_type = begin; @data_types.next; rescue StopIteration; nil; end
```
sdjmchattie, line 65: This feels a bit clunky to me.  A clever Ruby trick, but (I think) the block below is applied to the result of `.next` in the middle of this line and that's not clear enough IMHO.  Especially with the use of `;` as statement separators.

I would probably prefer to see the error being handled more explicitly across multiple lines.

sdjmchattie (changes requested):

PR #979, opened 2023-04-11
963 dpl 707 as developers we need to create the smrt link 12 revio version options sample sheet headers so users can have the correct columns on the sample sheet

enegoktan:
Closes #

*
*
* ...


config/pacbio_smrt_link_versions.yml
```diff
@@ -149,6 +154,51 @@ default: &default
       versions:
         - v10
         - v11
```
BenTopping, line 156: I would have a look to see if any of the existing values also apply to v12. https://www.pacb.com/wp-content/uploads/SMRT_Link_User_Guide_v12.0.pdf

config/pipelines/pacbio.yml
```diff
@@ -423,6 +423,7 @@ default: &default
               for:
                 - :well
               with: :well
+      # v12
```
BenTopping, line 426: assume this is going to be done in DPL-710

spec/models/pacbio/well_spec.rb
```diff
@@ -317,5 +324,77 @@
...
+            expect(well).to be_valid
+          end
+
+          well.include_base_kinetics = 'junk'
+          expect(well).not_to be_valid
+        end
+      end
+
+      context 'Library concentration' do
+        it 'must have a library concentration' do
```
BenTopping, line 386: Would add a test here and in the 'polymerase kit' tests to check it is valid when it has a valid value

config/pipelines/pacbio.yml
enegoktan: yes, that's the sample sheet config story

config/pacbio_smrt_link_versions.yml
enegoktan: Just had another look, don't think there's any other fields that are applicable to v12_revio

BenTopping (approved):

PR #1067, opened 2023-07-24
DPL-831: Refactor sample sheets to use existing messaging config structure

### LINKED ISSUE #1058: DPL-831: Refactor sample sheets to better handle multiple versions, instruments, and libraries
**Describe the Housekeeping**
Refactor sample sheets to better handle multiple versions, instruments, and libraries.

**Blocking issues**
- #1067
- #1085 

**Additional context**
As a follow-on to #993, the proposed solution (#1050) leaves room for improvement to maintain the quality of code going forward. An alternative implementation (#1057) provides additional ideas, but more work is required to form a simpler and more elegant solution.

This has resulted in #1067, part 1 - sample sheets to address the creation of the CSV.  
Part 2 of the solution addresses the instrument and version conflicts introduced by SMRT Link v12 for Sequel IIe.

StephenHulme:
Closes #1058 (part 1)

- Refactor sample sheets to use existing messaging config structure
- Use a multi-stage abstraction to compile down the Data -> JSON -> CSV
- Refactor the config parsing into a common class for both messaging and sample sheets
- CSVs return a nil field instead of an empty string for nil data
- Update documentation

Dependencies:
- #1091

Sample Sheet Changes:

[sample-sheets-diff.pdf](https://github.com/sanger/traction-service/files/12464051/sample-sheets-diff.pdf)

created with:  
_./scripts/download-sample-sheets $TRACTION_DEV_HOST 1 8 new-style_  
_git diff --no-index --word-diff=color --word-diff-regex=. storage/deprecated storage/new-style --color | [ansi2html](http://www.pixelbeat.org/scripts/ansi2html.sh) | [wkhtmltopdf](https://wkhtmltopdf.org/) - sample-sheets-diff.pdf_

### DIFF
--- a/app/exchanges/README.md
+++ b/app/exchanges/README.md
@@ -1,10 +1,12 @@
-# Sample Sheet Generator
+# Exchanges Output Generator
 
-The sample sheets are defined through configuration in
-[/config/pipelines/](/config/pipelines/)`{pipeline}.yml` for ONT and PacBio.
+The exchanges codebase generates messages and sample sheets for external interfaces.
 
-This configuration is parsed using
-[/app/exchanges/run_csv/](/app/exchanges/run_csv/)`{pipeline}_sample_sheet.rb` as appropriate.
+The messages and sample sheets are defined through configuration in
+[/config/pipelines/](/config/pipelines/)`{pipeline}.yml` for ONT and PacBio. The
+[DataStructureBuilder](/app/exchanges/data_structure_builder.rb) parses the configuration with
+additional functionality added to
+[/app/exchanges/run_csv/](/app/exchanges/run_csv/)`{pipeline}_sample_sheet.rb` as required.
 
 The required format of the sample sheets can be found in the SMRT Link User Guides
 (https://www.pacb.com/support/documentation/).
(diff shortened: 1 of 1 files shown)


app/csv_generator/pacbio_sample_sheet_message.rb
```diff
@@ -0,0 +1,33 @@
...
+  # eg ['System Name', 'Run Name']
+  def csv_headers
+    configuration.column_order
+  end
+
+  # Data Structure as csv
+  def payload
+    CSV.generate do |csv|
+      csv << csv_headers
+      # data_structure = {"sorted_wells"=>[{"Library Type"=>"Standard",
```
stevieing, line 15: This is recursive so maybe you can turn it into a recursive function? As long as you have a terminator it should be fine? It would also make Robocop happy.

config/pipelines/pacbio.yml
```diff
@@ -560,6 +560,100 @@ default: &default
                 - :well
                 - :sample
               with: :row_type
+        column_order:
```
stevieing, line 563: I suppose we could also add order as an attribute for each field but not sure which is best?

app/models/concerns/sample_sheet.rb
```diff
@@ -4,8 +4,30 @@
...
   module Well
+    # The sequencing_kit_box_barcode of this plate if well belongs to plate 1
+    # used for 2-plate sample sheets
+    def sequencing_kit_box_barcode_plate_1
+      plate.sequencing_kit_box_barcode if plate.plate_number == 1
+    end
+
+    # The sequencing_kit_box_barcode of this plate if well belongs to plate 2
+    # used for 2-plate sample sheets
+    def sequencing_kit_box_barcode_plate_2
```
stevieing, line 27: might be worth waiting until we merge develop but maybe we could have a method with an argument in the configuration. That would make it much more dynamic. Then for plate 1 and 2 we could have the same method but with a different argument?

app/exchanges/README.md
```diff
@@ -0,0 +1,75 @@
+# Sample Sheet Generator
+
+The sample sheets are defined through configuration in
+[/config/pipelines/](/config/pipelines/)`{pipeline}.yml` for ONT and PacBio.
+
+This configuration is parsed using
+[/app/exchanges/run_csv/](/app/exchanges/run_csv/)`{pipeline}_sample_sheet.rb` as appropriate.
+
+The required format of the sample sheets can be found on page 27 of the [SMRT Link User Guide
```
BenTopping, line 9: I'd make this less specific as we don't support just v12. Maybe something like 'The required format of the sample sheets can be found in the SMRT Link User Guide' (https://www.pacb.com/support/documentation/)

app/exchanges/README.md
```diff
@@ -0,0 +1,75 @@
...
+v12.0](https://www.pacb.com/wp-content/uploads/SMRT_Link_User_Guide_v12.0.pdf).
+
+Once generated, the sample sheets are uploaded to SMRT Link, a test instance can be found at
+[trac-smrtlink-uat.psd.sanger.ac.uk](https://trac-smrtlink-uat.psd.sanger.ac.uk:8243/sl/runs).
```
BenTopping, line 13: I tend to not include internal addresses in the repo's if I can avoid it just as a precautionary measure.

app/exchanges/message/message.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Message
+  # Message::Message
+  # Creates a message in the correct structure for the warehouse
+  class Message < DataStructureBuilder
+    # Produces the message in the correct format
+    # Example:
+    #   {"lims"=>"Traction", "bmap_flowcell"=>{"sample_uuid"=>"5",
```
BenTopping, line 9: No sure this example provides much help here.

app/exchanges/run_csv/deprecated_pacbio_sample_sheet.rb
BenTopping, line 1: Is this needed? If the issue is the structure of the config for v11 sample sheet, I would change the sample sheet config to work with the new method rather than support the old method if possible.

BenTopping: Agreed v10 and v11 support for new generation methods can be moved to a separate story / pr to avoid scope creep

BenTopping (changes requested):
I haven't pulled down locally to check yet. Just a few comments around documentation mainly.

BenTopping (approved):

PR #1081, opened 2023-08-01
Rename seed run names to be more descriptive

StephenHulme:
The sample runs used to seed a development instance of the application are generated with
```sh
bundle exec rails db:reset
bundle exec rails pacbio_data:create
```
The different runs have different configurations and properties, but it is not clear from the Runs page which run is which.

####  Changes proposed in this pull request

- Rename seed run names to be more descriptive
- Remove duplicated generated runs
- Major refactoring of sample run generation
- Minor code styling
- Update data so that seeded runs are accepted in SMRT-Link v12
  - Note that untagged libraries seem to cause an error regarding de-multiplexing in v12

Screenshot 2023-08-03 at 16 44 52


lib/tasks/pacbio_data.rake
```diff
@@ -28,17 +28,15 @@ namespace :pacbio_data do
...
-    pools = [
-      { library_type: 'Pacbio_HiFi', tag_set: nil, size: 1 },
-      { library_type: 'Pacbio_HiFi', tag_set: 'Sequel_16_barcodes_v3', size: 1 },
-      { library_type: 'Pacbio_HiFi_mplx', tag_set: 'Sequel_16_barcodes_v3', size: 5 },
-      { library_type: 'PacBio_IsoSeq_mplx', tag_set: 'IsoSeq_Primers_12_Barcodes_v1', size: 1 },
-      { library_type: 'Pacbio_IsoSeq', tag_set: 'IsoSeq_Primers_12_Barcodes_v1', size: 5 },
-      { library_type: 'Pacbio_IsoSeq', tag_set: 'IsoSeq_Primers_12_Barcodes_v1', size: 5 }
+    tag_sets = [
+      {  tag_set: nil, size: 1 },
+      {  tag_set: 'Sequel_16_barcodes_v3', size: 1 },
```
BenTopping, line 33: I wonder if this library_type was supposed to be associated with each request in a pool. Currently this value isn't used as library_type is generated in the reception_generator so it makes sense to remove it here. But the reception_generator gives the requests random library_types so it may not be as close to real data. Just worth noting.

BenTopping: Though these aren't pools they also aren't really tagsets as the size is used to determine how many libraries/pools are made from each set. Think it might be better to remove size and just cycle through each tagset a random number of times e.g. 1..5. Thoughts?

BenTopping (changes requested):
This rake task is a bit frankenstein-y because of the number of architecture changes we have made over time. Think the way we generate data could be reworked completely at some point  but I think this is a good improvement.

lib/tasks/pacbio_data.rake
```diff
@@ -15,52 +15,44 @@ namespace :pacbio_data do
...
-                                      insert_size: lib.insert_size,
-                                      libraries: [lib])
-        end.pool
-      end
+    # Generate a psuedo-unique barcode
+    # default => 12345678
+    # length: 4 => 1234
+    # bits: 36 => uur0cj2h
+    # Inspired by http://web.archive.org/web/20120925034700/http://blog.logeek.fr/2009/7/2/creating-small-unique-tokens-in-ruby
+    def barcode(length: 8, bits: 10)
```
BenTopping, line 34: 👍

BenTopping (changes requested):
Some minor comments. We probably want more pools/libraries for testing purposes, currently we only generate 2 of each with this script.

BenTopping (approved):

PR #1191, opened 2024-01-19
DPL-1000 - Support pacbio request aliquots

### LINKED ISSUE #1159: DPL-1000 - As developers we need to modify requests to use aliquots to support a more flexible way of doing multiplexing
**User story**
As developers we need to modify requests to use aliquots to support a more flexible way of doing multiplexing

James / Steve

PSD

**Acceptance criteria**
- [ ] A request can have many aliquots
- [ ] Reception creates primary aliquots for all PacBio requests imported.
- [ ] all existing requests should have a primary aliquot.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #1137
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1159

- Adds rake task to add primary aliquots to requests
- Adds aliquotable concern to requests
- Adds conditional to volume, concentration, template_prep_kit_box_barcode and insert_size presence validation on aliquot to not run for Pacbio::Request aliquots.

### DIFF
--- a/app/models/aliquot.rb
+++ b/app/models/aliquot.rb
@@ -16,9 +16,9 @@ class Aliquot < ApplicationRecord
   # currently I have set these to be validated but not sure
   # as library only validates when a run is created
   # maybe we need to do this when the state is set to used?
-  # we also need to consider that aliquot can be attached to a sample
-  # which will not have a barcode or insert size
-  # validates :volume, :concentration, :template_prep_kit_box_barcode, :insert_size, presence: true
+  # requests currently dont support these fields so we skip validation
+  validates :volume, :concentration, :template_prep_kit_box_barcode,
+            :insert_size, presence: true, unless: -> { source.is_a?(Pacbio::Request) }
   validates :volume, :concentration, :insert_size,
             numericality: { greater_than_or_equal_to: 0, allow_nil: true }
 end
--- a/app/models/aliquot.rb
+++ b/app/models/aliquot.rb
@@ -16,9 +16,10 @@ class Aliquot < ApplicationRecord
   # currently I have set these to be validated but not sure
   # as library only validates when a run is created
   # maybe we need to do this when the state is set to used?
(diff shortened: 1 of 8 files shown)


spec/models/aliquot_spec.rb
```diff
@@ -3,21 +3,21 @@
 require 'rails_helper'
 
 RSpec.describe Aliquot do
-  it 'is invalid without volume' do
-    expect(build(:aliquot, volume: nil)).not_to be_valid
-  end
+  # it 'is invalid without volume' do
```
stevieing, line 6: I think it would be better to set these tests to pending with a reason.

app/models/aliquot.rb
```diff
@@ -18,7 +18,7 @@ class Aliquot < ApplicationRecord
   # maybe we need to do this when the state is set to used?
   # we also need to consider that aliquot can be attached to a sample
   # which will not have a barcode or insert size
-  validates :volume, :concentration, :template_prep_kit_box_barcode, :insert_size, presence: true
+  # validates :volume, :concentration, :template_prep_kit_box_barcode, :insert_size, presence: true
```
stevieing, line 21: Is this the best way to do this?

stevieing (changes requested):

app/models/aliquot.rb
BenTopping: Yeah I did wonder if we want to only validate them if source is not on a request until (if) we add request volume tracking. Would that be better?

stevieing (approved):
Just one question. Will we create a rake task to run all of the migrations together?

BenTopping:
> Just one question. Will we create a rake task to run all of the migrations together?
A single rake task probably isn't needed.
We will do this one on its own because the other migrations rely on requests having a primary aliquot so this one can be run first and doesn't change anything user facing.

PR #1215, opened 2024-02-09
DPL-1066 - Modifies the pacbio library endpoint to only return libraries without pools.

BenTopping:

- Adds PacBio LibraryPoolResource. This is so the Pacbio PoolResource can access all PacBio Libraries.
- Updates PacBio LibraryResource to only return records without pools (new libraries).

One alternative approach is to this is to create a separate endpoint for the libraries without pools. I don't feel strongly about either approach as they both have their own issues.

Another approach is to filter out libraries with pools in the front end. However this will break pagination completely because the service and UI will be out of sync. e.g. service sends 25 libraries (some with pools), Ui only shows libraries without pools e.g. 10 libraries.

### DIFF
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -33,7 +33,7 @@ module V1
       end
 
       def self.records(options = {})
-        # If we go directly to the library resource we want to include only libraries withput pools
+        # If we go directly to the library resource we want to include only libraries without pools
         # Otherwise we only want to return all libraries
         if options[:include_directives]
            .instance_variable_get(:@resource_klass) == V1::Pacbio::LibraryResource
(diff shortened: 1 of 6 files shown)


app/resources/v1/pacbio/library_resource.rb
```diff
@@ -32,6 +31,12 @@ def self.default_sort
         [{ field: 'created_at', direction: :desc }]
       end
 
+      def self.records(_options = {})
+        # We only want to include only libraries without pools
+        # Libraries with pools should be referenced via the LibraryPoolResource
+        _model_class.where(pool: nil)
```
stevieing, line 37: I looked in other places where we have done similar and we use model_name (not sure if it is needed) and then super e.g. https://github.com/sanger/traction-service/blob/develop/app/resources/v1/pacbio/plate_resource.rb#L21. I wonder if this would be better than using _model_class as that is an internal method. We could also use a scope but it feels like there is no need for that as it is a temporary method.

BenTopping: Good spot, I have updated it.

stevieing (approved):
Nice. Looks really clean now.

PR #1221, opened 2024-02-19
DPL-1094 Cost code and study report view

### LINKED ISSUE #1217: DPL-1094 - As an SSR (Alice) I would like to run a regular report on the cost code and study information for ToL samples which can be incorporated into the google sheet for billing purposes
**User story**
As an SSR (Alice) I would like to run a regular report on the cost code and study information for ToL samples which can be incorporated into the google sheet for billing purposes

Alice / James / Stuart

Alice / James

**Acceptance criteria**
- [ ] For each library I would like to be able to retrieve the cost code from traction
- [ ] For each library I would like to retrieve the study uuid from traction
- [ ] I would like to link the study uuid to the study information in the warehouse
- [ ] I would like a read only account for traction service so I have the permissions to run the report
- [ ] I would like to have a view so I can retrieve all of this information in one go without having to create an SQL query
- [ ] I would like to be able to change the time periods so I can retrieve this information at regular intervals
- [ ] I would like to be able to incorporate this information easliy into the long read google sheet

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
This is an interim solution to get this work across the line. I would propose that we would turn this into a user interface and eventually look at automating as part of the long read automation project.

sdjmchattie:
Closes #1217 

- Add a view to access the data Alice needs.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -28,6 +28,9 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Preparing database =="
   system! 'bin/rails db:prepare'
 
+  puts "\n== Adding TOL tubes_report view =="
+  system! 'bin/rails db:migrate:redo VERSION=20240219180656'
+
   puts "\n== Removing old logs and tempfiles =="
   system! 'bin/rails log:clear tmp:clear'
 
--- a/bin/setup
+++ b/bin/setup
@@ -28,9 +28,6 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Preparing database =="
   system! 'bin/rails db:prepare'
 
-  puts "\n== Adding TOL tubes_report view =="
-  system! 'bin/rails db:migrate:redo VERSION=20240219180656'
-
   puts "\n== Removing old logs and tempfiles =="
   system! 'bin/rails log:clear tmp:clear'
 
--- a/lib/tasks/tol_tubes_report_view.rake
+++ b/lib/tasks/tol_tubes_report_view.rake
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+namespace :tol_tubes_report_view do
+  desc 'Manage the creation and destruction of the tubes_report view.'
+  task create: [:environment] do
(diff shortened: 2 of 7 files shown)


BenTopping (approved):
Looks good. At alternative approach could be to make a 'views' rake task and add this in there. Then add that rake task to the seeds.rb file so the views get generated on db:reset.

bin/setup
```diff
@@ -28,6 +28,9 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Preparing database =="
   system! 'bin/rails db:prepare'
 
+  puts "\n== Adding TOL tubes_report view =="
```
BenTopping: Can remove

lib/tasks/tol_tubes_report_view.rake
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+namespace :tol_tubes_report_view do
```
BenTopping, line 3: Did you look at whether we could do rspec tests on this? We have a few rake task specs to base it on but I understand this is slightly different because of the raw sql.

BenTopping (changes requested):

bin/setup
BenTopping: Fixed it before I published the comment. Time machine⏳

lib/tasks/tol_tubes_report_view.rake
sdjmchattie: I investigated adding tests and hit a few road blocks with testing the actual view data.  As we discussed, it's looking unlikely that we can test the data without some test packages being modified which is a bit of an undertaking for a view that's supposed to be quickly implemented.  When the test creates or destroys the view, the transaction that would otherwise be applied to the test data is overridden and commited to the database permanently across tests and test runs.  One solution is DatabaseCleaner but I think that's too far for this issue.

BenTopping (approved):

PR #1230, opened 2024-02-23
DPL-1109: Remove old API schema

### LINKED ISSUE #1224: DPL-1109 Remove the old schema Reception V1 API
**User story**
All clients of the Traction Service API are using the newer schema for Reception of data.  The old schema, which included request attributes, for example, can be removed to reduce technical debt.

Steve Inglis

Stuart McHattie

**Acceptance criteria**
- [x] Old fields used for the V1 schema have been removed from the API processing.
- [x] Any handling of optional V2 schema fields should be made mandatory.
- [ ] Test are up to date with support for V2 only.

sdjmchattie:
Closes #1224 

- Remove the old v1 API for receptions.
- Update tests to match the remaining behaviour.


spec/models/reception_spec.rb
```diff
@@ -36,4 +36,118 @@
...
     it { is_expected.not_to be_valid }
   end
+
+  context 'without plate_attributes or tubes_attributes' do
+    let(:attributes) { { plates_attributes: [], tubes_attributes: [] } }
+
+    it { is_expected.not_to be_valid }
+  end
+
+  describe '#construct_resources!' do
```
BenTopping, line 46: We already test construct_resources extensively in the resource factory so this can just be a basic test. We just want to test the association works so something like `expect reception.construct_resources! to change reception.requests by x`

BenTopping (changes requested):

stevieing (approved):
Satisfying to remove 900 lines of code.

spec/models/reception_spec.rb
sdjmchattie: Fair enough.  I thought all the other stuff was necessary for the resource test to work correctly, but I've reduced it right down now.

BenTopping (approved):

PR #1234, opened 2024-02-26
DPL-1110: Store genome size

### LINKED ISSUE #1225: DPL-1110 Process genome size as part of sample reception
**User story**
The samples received via the V1 Reception API can contain genome size as one of the fields.  This data should be written to the samples in the Traction database rather than being ignored.

Steve Inglis, James Watts

James Watts

**Acceptance criteria**
- [x] Samples with a genome size field have the genome size written to the Traction database.
- [x] Tests are updated to demonstrate that the field is not ignored but also isn't mandatory.

sdjmchattie:
Closes #1225 

- Add a column for genome size in the samples table.
- Make genome_size be an acceptable field on samples in the JSON API.
- Put a made up genome_size value in the test object.

### DIFF
--- a/app/resources/v1/reception_resource.rb
+++ b/app/resources/v1/reception_resource.rb
@@ -60,7 +60,7 @@ module V1
     def permitted_sample_attributes
       %i[name external_id species study_uuid priority_level public_name
          sanger_sample_id supplier_name taxon_id donor_id country_of_origin
-         accession_number date_of_sample_collection]
+         accession_number date_of_sample_collection genome_size]
     end
 
     def publish_messages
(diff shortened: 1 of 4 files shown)


stevieing (changes requested):
Genome size should be on the request model. There is already an attribute on request estimate_of_gb_required which I believe is the same thing. The type might need altering. I will check if there is any data in there.

stevieing (commented):
I have checked prod and there are only 10 instances of it being filled so it should not be a major issue changing it.

sdjmchattie:
Opening a new PR

PR #1236, opened 2024-02-27
DPL-989 Modify pool creation for aliquots

### LINKED ISSUE #1153: DPL-989 - As developers we need to modify pool creation to use aliquots rather than libraries to support revio multiplexing
**User story**
As developers we need to modify pool creation to use aliquots rather than libraries to support revio multiplexing

James / Steve

PSD

**Acceptance criteria**
- [ ] A pool should have a primary aliquot which will be derived from the derived aliquots
- [ ] A pool should have many derived aliquots which will come from libraries or samples
- [ ] When a pool is updated the appropriate aliquots should be updated.
- [ ] the api endpoint for pool creation should support inclusion or creation of aliquots
- [ ] All existing pools (of more than one library) should be migrated to use aliquots (rake task)

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #1137
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1153 

- Adds syncing in Pool model so Libraries and Used aliquots match each other


app/models/pacbio/pool.rb
```diff
@@ -16,24 +16,52 @@ class Pool < ApplicationRecord
...
     validates :volume, :concentration,
               :insert_size, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
+    validates :libraries, presence: true
+    validates :used_aliquots, presence: true
+
+    # List of common attributes between libraries and used aliquots
+    COMMON_ATTRIBUTES = %w[volume concentration insert_size
+                           template_prep_kit_box_barcode tag_id].freeze
 
     def library_attributes=(library_options)
```
stevieing, line 32: library_attributes and used_aliquot_attributes are very similar. Maybe we can do a bit of abstraction. Also does this need to be wrapped in a transaction?

app/models/pacbio/pool.rb
```diff
@@ -46,18 +74,44 @@ def sequencing_runs
...
+      end.each(&:destroy)
+    end
+
+    def destroy_unused_aliquots
+      used_aliquots.filter do |aliquot|
+        libraries.pluck(:pacbio_request_id).exclude?(aliquot.source_id)
+      end.each(&:destroy)
+    end
+
     def update_library(attributes)
```
stevieing, line 89: Again this and the following method are very similar.

app/models/pacbio/pool.rb
```diff
@@ -46,18 +74,44 @@ def sequencing_runs
...
                        .tap { |l| l.update(attributes) }
     end
 
+    def update_used_aliquot(attributes)
+      id = attributes['id'].to_s
+      indexed_used_aliquots.fetch(id) { missing_used_aliquot(id) }
+                           .tap { |a| a.update(attributes) }
+    end
+
     def missing_library(id)
```
stevieing, line 101: Again these are pretty much the same.

app/models/pacbio/pool.rb
```diff
@@ -46,18 +74,44 @@ def sequencing_runs
...
-      raise ActiveRecord::RecordNotFound, "Pacbio request #{id} is not part of the pool"
+      raise ActiveRecord::RecordNotFound, "Pacbio library is not part of the pool #{id}"
+    end
+
+    def missing_used_aliquot(id)
+      raise ActiveRecord::RecordNotFound, "Aliquot is not part of the pool #{id}"
     end
 
     def indexed_libraries
```
stevieing, line 109: And again. Very similar. Worth looking at abstraction.

app/resources/v1/pacbio/pool_resource.rb
```diff
@@ -43,11 +48,22 @@ def library_attributes=(library_parameters)
         @model.library_attributes = library_parameters.map do |library|
```
stevieing, line 48: we are using a constant below but not here?

stevieing (changes requested):
A few things to look at. Happy to discuss.

lib/tasks/migrate_pacbio_aliquot_data.rake
```diff
@@ -85,6 +85,11 @@ namespace :pacbio_aliquot_data do
       # Skip if aliquots already exist (shouldn't happen but just in case)
       next if pool.aliquots.any?
 
+      # Set defaults if they don't exist
+      pool.volume = pool.volume || 0
+      pool.concentration = pool.concentration || 0
+      pool.template_prep_kit_box_barcode = pool.template_prep_kit_box_barcode || '033000000000000000000'
```
seenanair, line 91: If we are using the same default values across libraries and pools, would it make sense to have these in somewhere common to both?

seenanair (commented):
I have only added a few doubts which I couldn't understand.

spec/lib/tasks/migrate_pacbio_aliquot_data.rake_spec.rb
```diff
@@ -127,6 +127,12 @@
       wells = create_list(:pacbio_well, 5, pool_count: 2)
       pools = wells.map(&:pools).flatten
 
+      # Get rid of aliquots that were created by the factory
+      pools.each do |pool|
+        pool.primary_aliquot.destroy
+        pool.used_aliquots.destroy_all
+      end
```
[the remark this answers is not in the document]
BenTopping: The gem we use to create dummy data is called factory bot and you generate data by using factories. The factories in this case are already updated to support aliquots but since this rake task test wants data without aliquots we get rid of any aliquots that were previously created.

app/resources/v1/pacbio/pool_resource.rb
BenTopping: I abstracted aliquot attributes because we use it in more than one place as opposed to library attributes. But I will add a library attributes constant for consistency.

app/models/pacbio/pool.rb
BenTopping: Yes I was thinking this code is going to be removed soon anyway so I dont want to spend ages on it but Ill admit this is glaringly refactor-able so I will give it a pass over and see what is easily abstracted.

app/models/pacbio/pool.rb
BenTopping: Yeah they are similar but use different methods and directly affect different properties which makes it a bit awkward to refactor. They don't need to be in a transaction as when used in the resource they are automatically wrapped in one.

spec/lib/tasks/migrate_pacbio_aliquot_data.rake_spec.rb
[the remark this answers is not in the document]
seenanair: Ok, now I got it. Thank you for explaining the details.

app/resources/v1/pacbio/library_resource.rb
```diff
@@ -24,17 +27,26 @@ class LibraryResource < JSONAPI::Resource
...
       end
 
-      def self.records(_options = {})
+      def self.records(options = {})
         # We only want to include only libraries without pools
         # Libraries with pools should be referenced via the LibraryPoolResource
-        super.where(pool: nil)
+        # This breaks aliquot polymorphism so we only filter if the resource
+        # Is accessed directly
+        if options[:include_directives]
```
BenTopping, line 44: This isn't ideal because we added the LibraryPoolResource to avoid doing this. But this adds a filter to the aliquot polymorphism that wipes out pools and we can't change the type from 'libraries' to 'librarypools' so we have to do this.

stevieing: Will this go after we do the tidy up?

BenTopping: Yes

app/resources/v1/aliquot_resource.rb
```diff
@@ -0,0 +1,27 @@
...
+    has_one :used_by, polymorphic: true,
+                      polymorphic_types: %w[pacbio/requests pacbio/pools pacbio/libraries]
+
+    has_one :tag, always_include_optional_linkage_data: true
+
+    # Aliquot polymorphism support
+    # This fixes the polymorphic relationships in json-api resources
+    # as json-api resources underscores and pluralizes the type
+    def self.resource_klass_for(type)
+      type = type.underscore.pluralize.gsub('pacbio/', '')
```
stevieing, line 23: This is really good but the main thing is that we are specifying pacbio in the top level aliquot resource which reduces flexibility. I think I might have a look at how we can use metaprogramming or composition so we can put something in aliquotable.

The other thing I am not sure of is why you are removing pacbio from the type?

BenTopping: Happy for you to have a look into it.

Yeah its slightly odd but in reality we are calling this from the `v1/pacbio/aliquots` resource and in there 'requests' 'libraries' and 'pools' resolve to their local namespace resource. So the pacbio prefix isnt needed and prevents that pacbio::pacbio bug. We might be able to move the pacbio references to the `v1/pacbio/aliquots` resource if you think thats better.

PR #1239, opened 2024-02-28
DPL-1094-2: Improve tubes_report view

### LINKED ISSUE #1238: DPL-1094-2: Add additional fields to the tubes_report view
**User story**
Alice need additional fields in the tubes_report view to help associate libraries with source samples.

Stuart McHattie

Alice L.

**Acceptance criteria**
- [ ] Source barcode appears in entries in the view.
- [ ] Source name appears in entries in the view.

**Additional context**
Source container barcodes can only be discovered via the `container_materials` table which is not yet shown in the [PacBio Domain Model](https://github.com/sanger/traction-service/blob/develop/erd-pacbio.jpg).  However the sample name is obtainable via the sample associated with the request.

sdjmchattie:
Closes #1238 

A lot of extra joins to get the data, but the new columns are for:

- The source identifier (which might be plate barcode + well position or it might be tube barcode).
- Add the sample name.

Also rename the existing barcode column to destination_barcode.

My plan will be to deploy this as usual but then trigger the rake task manually.  I don't want to include another identical migration just to trigger the task a second time.  The `create` task overwrites the existing view.


stevieing (approved):
The traction service production deployment is being done tomorrow @2pm. I'm not sure but we may need to change the view because of the library changes. I guess you can check tomorrow.

sdjmchattie:
> The traction service production deployment is being done tomorrow @2pm. I'm not sure but we may need to change the view because of the library changes. I guess you can check tomorrow.

Thanks @stevieing yes.  I assumed some of these relationships might change a bit.  @BenTopping mentioned how the destination tube barcode will become part of the library table schema and so I would not need to join pools and related tubes.  I'll check it tomorrow afternoon.

PR #1258, opened 2024-03-13
DPL-1112 - Handle used_aliquots for run creation

### LINKED ISSUE #1226: DPL-1112 - Update run creation/edit to use well aliquots
**User story**
For the Revio multiplexing work we will associate libraries and pools to wells through aliquots. This will require updates to the PacBio run api so the data associations and data structure reflects this change.

Ben T, Steve I, Seena N

**Acceptance criteria**
- [ ] When updating a run it creates a aliquot for each pool/library in a well. This aliquot will have a source of library/pool and a used_by well.
- [ ] Design the PacBio run API structure to allow the well aliquots to be created and updated correctly according to the data sent through the API.
- [ ] Consider supporting both the current process of creating/updating a run and supporting the aliquot data. This would allow us to switch between them freely and safely and reduce risk.
- [ ] Add a reversible migration to add well aliquots for all existing pools/libraries in wells.

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1226 

- Overloads well pool_ids and library_ids method to also handle used_aliquots creation and deletion

### DIFF
--- a/app/models/pacbio/well.rb
+++ b/app/models/pacbio/well.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
+# ALIQUOT-CLEANUP
+# - Update references for libraries and pools to be through used_aliquots
+# - Update library_ids and pools_ids to only update used_aliquots
+
 module Pacbio
   # Pacbio::Well
   # A well can have many libraries
-  class Well < ApplicationRecord
+  class Well < ApplicationRecord # rubocop:disable Metrics/ClassLength
     GENERIC_KIT_BARCODE = 'Lxxxxx100938900123199'
 
     include Uuidable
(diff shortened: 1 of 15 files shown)


app/models/pacbio/well.rb
```diff
@@ -52,6 +57,46 @@ class Well < ApplicationRecord
...
     delegate :run, to: :plate, allow_nil: true
 
+    def pool_ids=(ids)
+      ids.each do |id|
+        # If the used_aliquot already exists, skip it
+        next if used_aliquots.find_by(source_id: id)
+
+        # Create a new used_aliquot if it doesn't exist
+        used_aliquots << used_aliquots.build(source_id: id, source_type: 'Pacbio::Pool',
+                                             volume: 0, concentration: 0, aliquot_type: :derived,
```
BenTopping, line 67: I have given these a default for the time being since they need to pass validation. We can skip validation, not sure if thats something we want to be doing though.

lib/tasks/migrate_pacbio_aliquot_data.rake
```diff
@@ -126,6 +126,50 @@ namespace :pacbio_aliquot_data do
     end
   end
 
+  task migrate_well_data: :environment do
```
BenTopping, line 129: Do we want to add a primary aliquot for wells? I havent added one because I am not sure on the value it adds a well is not something that can be aliquoted itself.

stevieing: No. I don't think it makes sense either.

app/models/pacbio/well.rb
stevieing: I think this is an acceptable way to do this. When we come to do volume tracking we will need it.

stevieing (approved):

PR #1266, opened 2024-03-25
DPL-993 - Disallow deletion of libraries that are part of pools 

seenanair:
This PR includes
Added checks for  associated pools in a library before destroying a library

### DIFF
--- a/app/models/pacbio/library.rb
+++ b/app/models/pacbio/library.rb
@@ -45,7 +45,7 @@ module Pacbio
     accepts_nested_attributes_for :primary_aliquot
 
     after_create :create_used_aliquot, :create_tube, if: -> { pool.blank? }
-    before_destroy :check_for_associated_wells, prepend: true
+    before_destroy :check_for_associated_wells, :check_for_associated_pools, prepend: true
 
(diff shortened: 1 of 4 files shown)


app/models/pacbio/library.rb
```diff
@@ -101,5 +101,17 @@ def check_for_associated_wells
         throw(:abort)
       end
     end
+
+    def check_for_associated_pools
```
stevieing, line 105: Inwould think about turning this into a scope or a method in aliquot. That way everything is delegated to aliquot and you just need to return a boolean.

stevieing (changes requested):
I think this would be better as a method in Aliquot. You could pass an argument which would be type in this case Pacbio::Pool and then it can be reused.

BenTopping (approved):
Nice one 👍 
We should hold off merging until the current work is through but hopefully that will be in the next few days.

app/models/pacbio/library.rb
seenanair: 👍

stevieing (approved):

PR #1272, opened 2024-04-02
fix(ont-pools): fixes tag validation to go via libraries

BenTopping:
TagValidator is a shared class between PacBio and ONT. With the introduction of the aliquot work this has broken tag validation in ONT which does not have aliquots. 

- Updates TagValidator to only validate via used_aliquots if the record can respond to used_aliquots.

### DIFF
--- a/app/validators/tag_validator.rb
+++ b/app/validators/tag_validator.rb
@@ -6,7 +6,7 @@
 # if there is more than one library each tag will need to be unique
 class TagValidator < ActiveModel::Validator
   def validate(record)
-    if Flipper.enabled?(:multiplexing_phase_2_aliquot)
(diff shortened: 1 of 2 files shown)


seenanair (approved):
Looks good 👍 
Just one doubt, do we need any test case to check this?

BenTopping:
@seenanair Initially I thought so as I assumed we didnt have test coverage for this issue. However it is caused by the feature flag not being on in these tests so it uses the libraries method correctly and doesn't have any tests for when the feature flag is on. 
Not really sure how to catch this in the future. Maybe it would make sense to run the test suite twice, once with the flags on globally and once with them off.

PR #1274, opened 2024-04-03
DPL-998-1 - Cleanup aliquot/revio multiplexing work

### LINKED ISSUE #1156: DPL-998-1 - As developer we need to tidy up after the Revio multiplexing work to remove any code, data and tables that are no longer needed to reduce technical debt and improve maintainability
**Describe the Housekeeping**
As developer we need to tidy up after the Revio multiplexing work to remove any code, data and tables that are no longer needed to reduce technical debt and improve maintainability

**Blocking issues**
Describe any other issues or tickets that may be blocking this change.

**Additional context**
- sanger/traction-service#1137

Things to do:

- [ ] remove any relationships that are no longer needed e.g. library to pool
- [ ] remove any tables that are no longer needed e.g. well_libraries, well_pools
- [ ] tidy up the well validator to remove the need for validations that rely on there being more than 1 library or pool.
- [ ] Remove feature flags in the Service
- [ ] Remove transitional code that support both new/old api parameters in the service (libraries, pools, wells)
- [ ] Ensure library deletion hooks are moved over to check via aliquots.

BenTopping:
Closes #1156 

- Removes relationship between PacBio pools and libraries.
- Updates test references to pools with libraries.
- Removes feature flags.
- Removes PacBio WellPool and WellLibrary and updates well relationship for libraries and pools to go through used_aliquots.
- Updates WellValidator to remove redundant validations.
- Adds MultiSourcedPool to include library barcodes in the source_identifer method for PacBio pools.
- Updates TOL view in accordance to db changes.

#### Additional thoughts / questions
- PacBio has a concern called DualSourcedPool used to calculate the sources of the pool. Do we need to update this to something like MultiSourcedPool and include libraries as a source or do we always want the root source in which we could add library requests to the source identifier.


BenTopping:
05/04/24
Current failing test is caused by MySQL TOL view. Discussed with Stuart and decided to put it on hold until the TOL view has been finished in order to prevent unnecessary work reworking the view for aliquots.

app/models/pacbio/well.rb
```diff
@@ -60,12 +54,6 @@ class Well < ApplicationRecord # rubocop:disable Metrics/ClassLength
     delegate :run, to: :plate, allow_nil: true
 
     def pool_ids=(ids)
```
BenTopping, line 56: I think pool_ids an library_ids should be replaced by sending the used_aliquots directly. As now pools and libraries are no longer directly associated this doesn't really make sense. This would require changes to the UI as well which is why I haven't changed it yet. Probably worth doing when we do the well volume tracking work is done as that will require aliquots in the UI.

lib/tasks/tol_tubes_report_view.rake
```diff
@@ -70,4 +70,52 @@ namespace :tol_tubes_report_view do
       puts '-> TOL tubes report view successfully created' unless Rails.env.test?
     end
   end
+
+  namespace :v2 do
```
sdjmchattie, line 74: I think this also needs to be called in a migration.  Similar to https://github.com/sanger/traction-service/blob/develop/db/migrate/20240219180656_create_cost_code_and_study_report_view.rb but with `up` calling `v2:create` and down calling `v1:create`

app/validators/tag_validator.rb
```diff
@@ -6,7 +6,9 @@
 # if there is more than one library each tag will need to be unique
 class TagValidator < ActiveModel::Validator
   def validate(record)
-    if Flipper.enabled?(:multiplexing_phase_2_aliquot) && record.respond_to?(:used_aliquots)
+    # Since this is shared between PacBio (with aliquots) and ONT (with libraries)
```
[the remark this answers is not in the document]
BenTopping: Hopefully not. I think TagValidator is used indirectly by other tests but maybe worth adding a specific TagValidator test to catch future issues. Main fix came in https://github.com/sanger/traction-service/pull/1272.

lib/tasks/tol_tubes_report_view.rake
BenTopping: I normally run these kind of rake tasks directly on the servers. Happy to add a migration for it if you think that is the correct practice. It is tied to the irreversible migration where pacbio_library loses its pool_id association, so maybe I should call it as part of that?

sdjmchattie: Generally I think it's right to run these Rake tasks directly on the servers, but this one is modifying the database.  Unfortunately Rails doesn't support view management natively, so we did it this way instead.  I think you should add it to your existing migration file since it's tied to those columns being modified.

db/migrate/20240405075558_remove_pacbio_pool_id_from_pacbio_library.rb
```diff
@@ -0,0 +1,21 @@
...
+        library.delete
+      end
+    end
+
+    # Remove the column from the table
+    remove_reference :pacbio_libraries, :pacbio_pool, null: true, foreign_key: true
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
sdjmchattie, line 19: I do understand why this is irreversible because of the data loss that would occur when moving `up`.  Would it be better to not make this irreversible and just arrange the tables back the other way again even if the data is missing, or is that not possible because there must be data in those columns?

app/models/pacbio/library.rb
```diff
@@ -61,51 +59,31 @@ def create_used_aliquot
...
-      end
-    end
-
-    def check_for_associated_pools
-      return true unless derived_aliquots.any? do |aliquot|
-                           aliquot.used_by_is_a?(Pacbio::Pool)
-                         end
+    # Derived aliquots indicate it has been used in a pool or run
+    def check_for_derived_aliquots
+      return true if derived_aliquots.empty?
```
seenanair, line 84: Is it possible to have a case where derived aliquots exists but it is not used by? I hope it is not and derived_aliquots will be cleaned once an existing used_by relation is removed.

app/models/pacbio/pool.rb
```diff
@@ -1,26 +1,18 @@
...
 
-    # This is dependent on the requests association, so needs to be included
+    has_many :wells, through: :derived_aliquots, source: :used_by, source_type: 'Pacbio::Well'
+    has_many :requests, through: :used_aliquots, source: :source, source_type: 'Pacbio::Request'
+    has_many :libraries, through: :used_aliquots, source: :source, source_type: 'Pacbio::Library'
+
+    # This is dependent on the requests and libraries associations, so needs to be included
     # after that is defined
-    include DualSourcedPool
+    include MultiSourcedPool
```
seenanair, line 15: I have some questions about MultiSourcePool,  mostly stemming from my ignorance about Rail concepts. I will ask you offline. :)

seenanair (approved):
Great work which completes the whole multiplexing series 👏

app/models/pacbio/library.rb
BenTopping: Good question. The derived aliquot should be deleted when the used_by container is deleted so this _**should**_ never happen.

sdjmchattie (approved):
I've checked the migration and report view stuff carefully which looks good.  I glanced over the rest and didn't see anything alarming, but you should probably wait for others to approve those changes.

sdjmchattie:
> I've checked the migration and report view stuff carefully which looks good. I glanced over the rest and didn't see anything alarming, but you should probably wait for others to approve those changes.

Looks like they already did!

app/models/pacbio/well.rb
BenTopping: https://github.com/sanger/traction-ui/issues/1518

PR #1275, opened 2024-04-04
Y24-018: Tube reception accepts library creation attributes

### LINKED ISSUE #1255: Y24-018 Validate that an API endpoint can place a compound sample tube as a library
**Description**
Messages coming via a queue consumer will need to be able to call an API endpoint to pass a Pool XP tube into Traction Service that was originated from Limber/SequenceScape.  We need to establish whether the existing API endpoints are capable of doing this, and if not we should add to the API so that it becomes possible

**Who the primary contacts are for this work**
Stuart McHattie

Stephen Inglis

We need to discuss how the tube will exist in Traction.  There is the concept of libraries, but the tube is likely to be representative of 9,124 samples without passing the actual samples over to Traction.  So it might be a tube with a simple sample in it rather than an actual library.  This should be discussed with Stephen Inglis and maybe Ben Topping.

sdjmchattie:
Closes #1255 

- Reception message accepts data for library attributes against tubes.
- When these are missing, the result is the same as before.
- When these are there and valid, the library is created for the request.
- When these are there but invalid, the service gives a 422 response.

### DIFF
--- a/app/models/reception/resource_factory.rb
+++ b/app/models/reception/resource_factory.rb
@@ -29,6 +29,10 @@ class Reception
       @containers ||= []
     end
 
+    def libraries
+      @libraries ||= []
+    end
+
     def requests
       @requests ||= []
     end
@@ -38,6 +42,7 @@ class Reception
     end
 
     def construct_resources!
+      libraries.each(&:save!)
       requests.each(&:save!)
       labware_status
     end
@@ -117,8 +122,17 @@ class Reception
     def create_request_for_container(attributes, container)
       library_type = library_type_for(attributes[:request])
       sample = sample_for(attributes[:sample])
-      requests << create_request(library_type, sample, container, attributes[:request])
+      request = create_request(library_type, sample, container, attributes[:request])
+      if attributes[:library]
+        libraries << create_library(library_type, request, attributes[:library])
+      end
+      requests << request
       containers << container
+    end
+
+    def create_library(library_type, request, library_attributes)
+      # The library type is used to help build the correct library in the correct pipeline
+      library_type.library_factory(request:, library_attributes:)
     end
 
     def create_request(library_type, sample, container, request_attributes)
--- a/app/models/reception/resource_factory.rb
+++ b/app/models/reception/resource_factory.rb
@@ -79,6 +79,16 @@ class Reception
       end
     end
 
+    def library_type_for(request_attributes)
+      # Gets a library type from the cache or returns an unknown type
+      library_type = request_attributes[:library_type]
+      library_type_cache.fetch(
+        library_type,
+        UnknownLibraryType.new(library_type:,
+                               permitted: library_type_cache.keys)
+      )
+    end
+
     def data_type_for(attributes)
       data_type_cache.fetch(attributes[:data_type], nil)
     end
@@ -108,9 +118,11 @@ class Reception
     end
 
     def create_request_for_container(attributes, container)
-      library_type_helper = LibraryTypeHelper.new(self, attributes)
+      library_type_helper = LibraryTypeHelper.new(library_type_for(attributes[:request]),
+                                                  self,
+                                                  attributes)
       sample = sample_for(attributes[:sample])
-      request = library_type_helper.create_request(sample, container)
+      request = library_type_helper.create_request(sample, container, reception)
       library_type_helper.create_library(request)
       requests << request
       containers << container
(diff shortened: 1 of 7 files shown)


sdjmchattie:
Missing coverage in tests is related to pipelines we haven't intentionally tried to add support for.  Best efforts were made to ensure the framework is there if we do want to support it in future.

app/ont/ont.rb
```diff
@@ -6,6 +6,10 @@ def self.table_name_prefix
     'ont_'
   end
 
+  def self.library_attributes
+    []
```
BenTopping, line 10: I believe this would be `volume, concentration, insert_size, tag_id`. Might be better just throwing an error if this method is reached as unsupported. e.g. `raise StandardError, 'Unsupported'`

app/ont/ont.rb
```diff
@@ -20,6 +24,13 @@ def self.direct_request_attributes
     request_attributes - associated_request_attributes
   end
 
+  def self.library_factory(request:, library_attributes:)
```
BenTopping, line 27: Same here I think this should be `raise StandardError, 'Unsupported'` for the time being and then we support it if there becomes a need for it.

app/resources/v1/reception_resource.rb
```diff
@@ -53,6 +54,10 @@ def construct_resources!
       context[:labware] = @model.construct_resources!
     end
 
+    def permitted_library_attributes
```
BenTopping, line 57: Only supporting PacBio so can remove ONT and Saphyr

app/saphyr/saphyr.rb
```diff
@@ -6,6 +6,10 @@ def self.table_name_prefix
     'saphyr_'
   end
 
+  def self.library_attributes
+    []
```
BenTopping, line 10: Same comments as ONT. This and library factory should probably raise an unsupported error.

app/pacbio/pacbio.rb
BenTopping, line 1: Can't see a unit test file for this, wonder if its worth adding one.

spec/requests/v1/reception_spec.rb
```diff
@@ -436,4 +436,140 @@ def publish(message)
...
+          expect(new_library.insert_size).to be_nil
+          expect(new_library.template_prep_kit_box_barcode).to eq('barcode')
+        end
+      end
+
+      context 'library is missing volume' do
+        let(:library) { { concentration: 2, template_prep_kit_box_barcode: 'barcode' } }
+
+        it 'responds with correct result' do
+          post v1_receptions_path, params: body, headers: json_api_headers
```
BenTopping, line 551: We should check the error message is as we expect as well

spec/requests/v1/reception_spec.rb
```diff
@@ -436,4 +436,140 @@ def publish(message)
...
+      context 'library is missing template prep kit box barcode' do
+        let(:library) { { volume: 1, concentration: 2 } }
+
+        it 'responds with correct result' do
+          post v1_receptions_path, params: body, headers: json_api_headers
+          expect(response).to have_http_status(:unprocessable_entity), response.body
+        end
+      end
+    end
+  end
```
BenTopping, line 574: Another test for library missing all attributes.

app/models/reception/resource_factory.rb
```diff
@@ -148,5 +144,37 @@ def duplicate_containers
...
+  class LibraryTypeHelper
+    def initialize(library_type, resource_factory, container_attributes)
+      @library_type = library_type
+      @resource_factory = resource_factory
+      @container_attributes = container_attributes
+    end
+
+    def create_library(request)
+      # The library type is used to help build the correct library in the correct pipeline
+      return unless @container_attributes[:library]
```
BenTopping, line 164: Would this return if we send an empty library object?

BenTopping (changes requested):
Couple minor comments mainly about other pipeline support.

spec/requests/v1/reception_spec.rb
```diff
@@ -436,4 +436,140 @@ def publish(message)
...
+        let(:library) { { volume: 1, concentration: 2, insert_size: 3, template_prep_kit_box_barcode: 'barcode' } }
+
+        it 'has a created status' do
+          post v1_receptions_path, params: body, headers: json_api_headers
+          expect(response).to have_http_status(:created), response.body
+        end
+
+        it 'creates the correct library for the request' do
+          expect { post v1_receptions_path, params: body, headers: json_api_headers }
+            .to change(Pacbio::Library, :count)
```
BenTopping, line 513: Also worth checking it creates the request properly.

BenTopping: And that the Pacbio::Library.request is the request that gets created.

app/models/reception/resource_factory.rb
sdjmchattie: No, it tries to create the library.  I'm writing the test for it doing that.  I think if someone's providing a `library` key they probably meant to create a library and not doing so silently would be confusing.

app/pacbio/pacbio.rb
sdjmchattie: It is tested implicitly by the `reception_spec.rb` file.  I had noticed it was missing the separate test file along with `ont.rb` and `saphyr.rb`.  I can create those if we want, but the files are pretty simple and, in my mind, more like helpers so that the correct types get generated for the selected pipeline.

BenTopping: Yes probably not worth the effort since they are tested as auxiliaries in other tests.

BenTopping (approved):

PR #1287, opened 2024-04-15
DPL-1069: Added volume methods and checks to support volume tracking

seenanair:
This PR includes
- added used_volume, available_volume and volume_check in `aliquotable` concern

### DIFF
--- a/app/models/concerns/aliquotable.rb
+++ b/app/models/concerns/aliquotable.rb
@@ -46,7 +46,6 @@ module Aliquotable
     # It subtracts the used volume from the volume of the primary aliquot.
     # @return [Numeric] The volume available in the primary aliquot after subtracting
     # the used volume.
-
     def available_volume
       primary_aliquot.volume - used_volume
     end
@@ -59,7 +58,6 @@ module Aliquotable
     # @return [Boolean] Returns true if the available volume is greater than
     # or equal to the required volume.
     # If not, it adds an error message to `errors` and returns false.
-
     def volume_check(required_volume)
       return true if available_volume >= required_volume
 
--- a/app/models/concerns/aliquotable.rb
+++ b/app/models/concerns/aliquotable.rb
@@ -35,17 +35,31 @@ module Aliquotable
              as: :source, class_name: 'Aliquot',
              dependent: :nullify, inverse_of: :source
 
-    # Method to calculate the used volume
+    # Method to calculate the used volume.
+    # It sums up the volumes of all derived aliquots.
+    # @return [Numeric] The total volume of all derived aliquots.
     def used_volume
       derived_aliquots.sum(&:volume)
     end
 
-    # Method to calculate the available volume
+    # Method to calculate the available volume.
+    # It subtracts the used volume from the volume of the primary aliquot.
+    # @return [Numeric] The volume available in the primary aliquot after subtracting
+    # the used volume.
+
     def available_volume
       primary_aliquot.volume - used_volume
     end
 
-    # Method to check if there is enough volume available
+    # Method to check if there is enough volume available.
+    # It compares the available volume with the required volume.
+    #
+    # @param required_volume [Numeric] The volume required for the operation.
+    #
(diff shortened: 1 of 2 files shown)


app/models/concerns/aliquotable.rb
```diff
@@ -34,5 +34,23 @@ module Aliquotable
     has_many :derived_aliquots, -> { where(aliquot_type: :derived) },
              as: :source, class_name: 'Aliquot',
              dependent: :nullify, inverse_of: :source
+
+    # Method to calculate the used volume
+    def used_volume
+      derived_aliquots.sum(:volume)
```
BenTopping, line 40: Im not sure how this will behave it derived_aliquots is empty. This may need to be `derived_aliquots.sum(&:volume)` See 'Safe Navigation Operator section' https://ruby-doc.org/core-2.6/doc/syntax/calling_methods_rdoc.html.

spec/models/concerns/aliquotable_spec.rb
```diff
@@ -48,4 +48,43 @@
       expect(pacbio_pool.used_aliquots).to eq []
     end
   end
+
+  describe '#used_volume' do
```
BenTopping, line 52: Additional test for when there are no derived_aliquots.

app/models/concerns/aliquotable.rb
seenanair: Thank you. Learned a new Ruby thing 👍

stevieing (changes requested):

app/models/concerns/aliquotable.rb
```diff
@@ -34,5 +34,23 @@ module Aliquotable
...
+    def used_volume
+      derived_aliquots.sum(&:volume)
+    end
+
+    # Method to calculate the available volume
+    def available_volume
+      primary_aliquot.volume - used_volume
+    end
+
+    # Method to check if there is enough volume available
```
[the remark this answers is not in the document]
seenanair: Ok, I will update it. 
According to the story, the aliquot volume is passed in. So, I think I will update it now with a volume and change if it aliquot need to be passed in. Is that ok?

app/models/concerns/aliquotable.rb
```diff
@@ -34,5 +34,23 @@ module Aliquotable
...
+    def volume_check(required_volume)
+      return true if available_volume >= required_volume
```
seenanair, line 50: @stevieing @BenTopping  I thought it make sense to return true in cases where the required_volume is equal to available_volume. Something for you to confirm.

stevieing: It might be that we need to provide a message that the volume is exhausted but we don't know until it gets used.

app/models/concerns/aliquotable.rb
```diff
@@ -34,5 +34,23 @@ module Aliquotable
     has_many :derived_aliquots, -> { where(aliquot_type: :derived) },
              as: :source, class_name: 'Aliquot',
              dependent: :nullify, inverse_of: :source
+
+    # Method to calculate the used volume
```
stevieing, line 38: I think a bit more documentation would be nice. e.g. the used_volume is the sum of all of the volumes of the derived aliqouts. Also adding parameters and return values as per the standard heading documentation.

app/models/concerns/aliquotable.rb
```diff
@@ -34,5 +34,23 @@ module Aliquotable
...
+    def used_volume
+      derived_aliquots.sum(&:volume)
+    end
+
+    # Method to calculate the available volume
+    def available_volume
```
stevieing, line 44: Same as above.

stevieing (changes requested):
I suspect we will need to make a few changes as we progress and get some feedback but we have enough for now.

Just a couple of changes for documentation.

stevieing (approved):
I think we can just merge this in? As it does not impact on anything else.

seenanair:
> I think we can just merge this in? As it does not impact on anything else.

Yes, I think so. 👍

PR #1288, opened 2024-04-16
DPL-1070: Volume check added when library volume is updated

seenanair:
This PR includes

- volume_check added in library modal when a library volume is modified

### DIFF
--- a/app/models/pacbio/library.rb
+++ b/app/models/pacbio/library.rb
@@ -43,6 +43,12 @@ module Pacbio
     accepts_nested_attributes_for :primary_aliquot
 
     after_create :create_used_aliquot
+
+    # Before updating the record, the `check_volume` method is called.
+    # This is a callback that checks if the volume has been changed and
+    # if the new volume is greater than the used volume.
+    before_update :check_volume
+
     before_destroy :check_for_derived_aliquots, prepend: true
 
     def create_used_aliquot
@@ -84,6 +90,23 @@ module Pacbio
       return true if derived_aliquots.empty?
 
       errors.add(:base, 'Cannot delete a library that is used in a pool or run')
+      throw(:abort)
+    end
+
+    # Checks if the volume has been changed and if the new volume is greater than the used volume.
+    # If the volume has not been changed, the method immediately returns.
+    # If the new volume is greater than the used volume, the method returns true.
+    # If the new volume is less than or equal to the used volume
+    # - an error is added to the volume attribute and
+    # - the update operation is aborted.
+    #
+    # @return [TrueClass, nil]
+    # Returns true if the new volume is greater than the used volume, nil otherwise.
+    def check_volume
+      return unless volume_changed?
+      return true if volume > used_volume
+
+      errors.add(:volume, 'Volume must be greater than the current used volume')
       throw(:abort)
     end
   end
--- a/spec/models/pacbio/library_spec.rb
+++ b/spec/models/pacbio/library_spec.rb
@@ -228,6 +228,47 @@ RSpec.describe Pacbio::Library, :pacbio do
     end
   end
 
+  describe 'when volume is updated' do
+    context 'when volume is updated to a value greater than used_volume' do
+      it 'allows the update' do
+        library = create(:pacbio_library, volume: 100)
+        create_list(:aliquot, 5, aliquot_type: :derived, source: library, volume: 2)
+        library.volume = 50
+        allow(library).to receive(:check_volume).and_return(true)
+        expect(library.save).to be_truthy
+      end
+    end
+
+    context 'when volume is updated to a value less than used_volume' do
+      let!(:library) { create(:pacbio_library, volume: 100) }
+
+      before do
+        create_list(:aliquot, 5, aliquot_type: :derived, source: library, volume: 2)
+        library.volume = 5
+      end
+
+      it 'does not allow the update' do
+        library.save
+        expect(library.save).to be_falsey
+      end
+
+      it 'adds an error message' do
+        library.save
+        expect(library.errors[:volume]).to include('Volume must be greater than the current used volume')
+      end
+    end
+
+    context 'when a field other than volume is updated' do
+      it 'allows the update' do
+        library = create(:pacbio_library, volume: 100)
+        create_list(:aliquot, 5, aliquot_type: :derived, source: library, volume: 2)
+        library.concentration = 50
+        expect(library.save).to be_truthy
+        expect(library.errors[:volume]).to be_empty
+      end
+    end
+  end
+
   describe '#sequencing_plates' do
     it 'when there is no run' do
       library = create(:pacbio_library)


app/models/pacbio/library.rb
```diff
@@ -86,5 +92,22 @@ def check_for_derived_aliquots
...
+    # Checks if the volume has been changed and if the new volume is greater than the used volume.
+    # If the volume has not been changed, the method immediately returns.
+    # If the new volume is greater than the used volume, the method returns true.
+    # If the new volume is less than or equal to the used volume
+    # - an error is added to the volume attribute and
+    # - the update operation is aborted.
+    #
+    # @return [TrueClass, nil]
+    # Returns true if the new volume is greater than the used volume, nil otherwise.
+    def check_volume
```
stevieing, line 105: would it be better to use the volume check method that you created in aliquotable?

app/models/pacbio/library.rb
```diff
@@ -43,6 +43,12 @@ class Library < ApplicationRecord
     accepts_nested_attributes_for :primary_aliquot
 
     after_create :create_used_aliquot
+
+    # Before updating the record, the `check_volume` method is called.
+    # This is a callback that checks if the volume has been changed and
+    # if the new volume is greater than the used volume.
+    before_update :check_volume
```
stevieing, line 50: Doing a before_update here is a good idea. We can check with long read to see if this is acceptable. It would also be worth checking whether we can add this to aliquotable. The more stuff we can put in the concern the better.

stevieing (changes requested):

app/models/pacbio/library.rb
seenanair: Please correct, if I am wrong.
I thought they are used for different purposes, where aliquotable volume check verifies whether the given volume is less than the available volume, and library volume check checks whether the given volume is less than the used volume.

app/models/pacbio/library.rb
```diff
@@ -86,5 +92,22 @@ def check_for_derived_aliquots
...
+    # If the new volume is greater than the used volume, the method returns true.
+    # If the new volume is less than or equal to the used volume
+    # - an error is added to the volume attribute and
+    # - the update operation is aborted.
+    #
+    # @return [TrueClass, nil]
+    # Returns true if the new volume is greater than the used volume, nil otherwise.
+    def check_volume
+      return unless volume_changed?
+      return true if volume > used_volume
```
BenTopping, line 107: I don't know whether we want to be checking the volume or the primary_aliquot volume. Currently the request from the UI will include and update both the library volume and primary_aliquot volume.

BenTopping (commented):
Changes look sensible.

I think we also need to think about how this is going to fit in to DPL-1075 (UI side). We need to decide how the volume check is going to work. What will the UX be - Is it going to error when users try and update? Do we warn them before they update it? Can they override the check and forcefully update the volume?

Continuing from the above we also need to think about historical data. I doubt the volume data will be accurate so users may need to update volume and ignore this check.

stevieing:
> Changes look sensible.
> 
> I think we also need to think about how this is going to fit in to DPL-1075 (UI side). We need to decide how the volume check is going to work. What will the UX be - Is it going to error when users try and update? Do we warn them before they update it? Can they override the check and forcefully update the volume?
> 
> Continuing from the above we also need to think about historical data. I doubt the existing volume data will be accurate so users may need to update volume and ignore this check.

I am seeing James Watts today so I was going to ask about this.

app/models/pacbio/library.rb
stevieing: I think we should check the primary_aliquot volume as this is what we will be using in the future. I presume we will eventually deprecate volume?

seenanair: The method changed to checking the primary aliquot volume

stevieing (approved):

BenTopping (approved):
Changes look good.
Just wondering about feature flags? Does it make sense to turn on this validation before the rest of the library volume checks are in place like library in pool and library in well.

app/models/concerns/aliquotable.rb
```diff
@@ -50,19 +50,49 @@ def available_volume
...
+    # the method returns true.
+    # 3. If the volume of the primary aliquot is less than the used volume,
+    # the method adds an error to the library record and aborts the update operation.
+    #
+    # @return [nil, true, false] Returns nil if the primary aliquot has not changed its volume,
+    #  true if the volume of the primary aliquot is greater than or equal to the used volume, and
+    # false if the volume of the primary aliquot is less than the used volume.
+
+    def primary_aliquot_volume_sufficient
+      return unless primary_aliquot&.volume_changed?
```
BenTopping, line 91: Not sure this check (line 91) should be here. If we add a used_aliquot but don't change the primary_aliquot we still want to check the primary aliquot has sufficient volume.

seenanair: I thought it will be covered by available_volume_sufficient method.

Please correct, if I am wrong. My understanding is

The `primary_aliquot_volume_sufficient` is called when library volume is updated/edited (attached to library model hook before_update :primary_aliquot_volume_sufficient) 
and  available_volume_sufficient is called when a used_aliquot value is updated (e.g when a library added  to pool)

seenanair:
> Changes look good. Just wondering about feature flags? Does it make sense to turn on this validation before the rest of the library volume checks are in place like library in pool and library in well.

Yes, perhaps.
This validation is actually implemented on the library edit page, independent of used volumes (both in the library pool and in wells). It ensures that the user can never edit a volume in an existing library to be less than the used volumes. 

The checks for used volume will occur when we add a library to a pool/well, ensuring that the used volume never exceeds the primary aliquot volume. 
These checks can exist independently. However, if we want to consolidate all volume checks into one step, we could potentially implement this using a feature flag.

app/models/concerns/aliquotable.rb
BenTopping: Ah yes good point. I think when a library is added to a pool we could call valid on the library as now it has the used_aliquot from the pool. Ill look into it in DPL-1072 and update accordingly.

seenanair: Yes, with the new changes, the concept of "primary_aliquot_volume_sufficient" is somewhat obsolete. 

Currently, the UI prevents the user from editing a value less than the primary aliquot volume, but the user experience aspect of this restriction needs to be explored further. It may limit user interaction, especially when typing into the volume text field. 
This is something you can also check as part of the review process, please.
 If this restriction proves to be problematic for the user experience, we may need to reconsider its implementation and potentially retain both functionalities `used_volume` and `primary_aliquot_volume_sufficient` check.

config/feature_flags.yml
```diff
@@ -1,4 +1,6 @@
 # Add feature flags to this file, example:
 # dpl202_enable_angry_cat_feature: When enabled this will add an angry cat to all pages that will try to catch the mouse pointer
 # The key should be lowecase and snake case, and prefixes with the issue number to allow easy identification
-{}
+{
```
BenTopping, line 4: don't think you need it wrapped in an object

BenTopping (approved):

PR #1291, opened 2024-04-17
DPL 1070 - aliquotable volume check

seenanair:
This story aims to split the PRs for volume checks in the 'Aliquotable' concern from the callback hook in the 'Library' model ([PR](https://github.com/sanger/traction-service/pull/1288)) 
This separation enables independent deployment of this story, whereas changes to the Library model necessitate more extensive testing.

### DIFF
--- a/app/models/concerns/aliquotable.rb
+++ b/app/models/concerns/aliquotable.rb
@@ -50,7 +50,7 @@ module Aliquotable
       primary_aliquot.volume - used_volume
     end
 
-    # Method: available_volume_check
+    # Method: is_available_volume_sufficient
     #
     # This method is used to validate if the available volume is sufficient for a required volume.
     #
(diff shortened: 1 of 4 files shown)


stevieing (approved):

app/models/concerns/aliquotable.rb
```diff
@@ -50,19 +50,50 @@ def available_volume
...
+    # The method performs the following checks:
+    # 1. If the available volume is greater than or equal to the required volume,
+    # the method returns true.
+    # 2. If the available volume is less than the required volume, the method adds an error to the
+    # base attribute of the record and returns false.
+    #
+    # @param required_volume [Numeric] The volume that is required.
+    #
+    # @return [Boolean] Returns true if the available volume is sufficient, false otherwise.
+    def available_volume_check(required_volume)
```
dasunpubudumal, line 66: Hi @seenanair, just suggesting here; does a method name like `is_available_volume_sufficient` convey the purpose better? Same with the `used_volume_check`.

seenanair: Yes, that would be more descriptive 👍 . I will change it to available_volume_sufficient (As RuboCop suggested to Rename `is_available_volume_sufficient` to `available_volume_sufficient?` ) and primary_aliquot_volume_sufficient (as we are checking it's sufficiency rather than used_volume)

dasunpubudumal (approved):

PR #1310, opened 2024-05-08
Y24-073 - Populate PacBio run comments with pool barcode + concentration

### LINKED ISSUE #627: Y24-073 - As a user I would like the run comments to be populated with the pool barcode and concentration for each well in the run to save time populating this manually
**User story**
As a user I would like the pool barcode and concentration for each well to be in the pacbio run comments. The required format is "PoolBarcode1 Concentration1pM PoolBarcode2 Concentration2pM PoolBarcode3 Concentration3pM ..." 

Craig

**Acceptance criteria**
- [x] automatically update the run comments when a run is updated/created with the correct pool barcodes and concentrations

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

SHIV5T3R:
Closes #627 

- When a run is created, the barcode will have the format: "BARCODE CONCENTRATION". The same will happen when a run is updated as long as the comment is empty.

### DIFF
--- a/app/models/pacbio/run.rb
+++ b/app/models/pacbio/run.rb
@@ -12,6 +12,11 @@ module Pacbio
     # Sequel II and Sequel I are now deprecated
     enum system_name: { 'Sequel II' => 0, 'Sequel I' => 1, 'Sequel IIe' => 2, 'Revio' => 3 }
 
+    # before_create :generate_comment, unless: -> { wells.nil? }
+
+    # We want to generate comments before the run was created
+    # but tube barcodes aren't generated until the run is created.
+
     after_create :generate_name, :generate_comment
 
     has_many :plates, foreign_key: :pacbio_run_id,
--- a/app/models/pacbio/run.rb
+++ b/app/models/pacbio/run.rb
@@ -12,7 +12,7 @@ module Pacbio
     # Sequel II and Sequel I are now deprecated
     enum system_name: { 'Sequel II' => 0, 'Sequel I' => 1, 'Sequel IIe' => 2, 'Revio' => 3 }
 
-    after_create :generate_name
+    after_create :generate_name, :generate_comment
 
     has_many :plates, foreign_key: :pacbio_run_id,
                       dependent: :destroy, inverse_of: :run, autosave: true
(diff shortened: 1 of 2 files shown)


app/models/pacbio/run.rb
```diff
@@ -46,7 +46,12 @@ class Run < ApplicationRecord
 
     # if comments are nil this blows up so add try.
     def comments
-      super || wells.try(:collect, &:summary).try(:join, ':')
+      return super if super.present?
+
+      wells.collect do |well|
+        "#{well.used_aliquots.first.source.tube.barcode} " \
+          "#{well.smrt_link_options['library_concentration']}pM"
```
stevieing, line 53: You can access library_concentration directly i.e `well.library_concentration`,

smrt_link_options are a store which provides accessors.

stevieing (changes requested):

stevieing (approved):

app/models/pacbio/run.rb
```diff
@@ -12,7 +12,7 @@ class Run < ApplicationRecord
     # Sequel II and Sequel I are now deprecated
     enum system_name: { 'Sequel II' => 0, 'Sequel I' => 1, 'Sequel IIe' => 2, 'Revio' => 3 }
 
-    after_create :generate_name
+    after_create :generate_name, :generate_comment
```
stevieing, line 15: I think it might make sense to do this as a before_create so we don't have to do 2 saves. Sorry my fault didn't think aboout this.

app/models/pacbio/run.rb
```diff
@@ -45,8 +45,20 @@ class Run < ApplicationRecord
     attr_reader :plates_attributes
 
     # if comments are nil this blows up so add try.
+
+    def generate_comment
```
stevieing, line 49: I think you can simplify this. 

```
comment = wells.collect do |well|
        " #{well.used_aliquots.first.source.tube.barcode} #{well.library_concentration}pM"
      end.join(' ')

this.comments += comment
```

app/models/pacbio/run.rb
```diff
@@ -45,8 +45,20 @@ class Run < ApplicationRecord
...
+      comment = "#{comments} ".presence || ''
```
stevieing, line 50: I am not sure you need to set comments to an emty string here as you are already doing this in comments accessor.

stevieing (changes requested):

SHIV5T3R:
I've left the last line as `update(comments: (comments + comment)[0, 65535])` because the comment seems not be updating with this change. The splicing is for one of the tests. Because it's appending the new format, it'll be over the limit for the database itself and because of the way the run is created, the comment can't be spliced in the test. Not really sure how to go about this part.

stevieing (approved):

PR #1314, opened 2024-05-14
DPL-1073 : Validate library volume when adding a library to a PacBio Run 

seenanair:
Changes proposed in this pull request

- Adds validation to PacBio wells to check libraries used have sufficient volume
- Adds feature flag dpl_1076_check_library_volume_in_runs to control volume validation for libraries in runs
- Replaced pool_ids and library_ids with accepts_nested_attributes_for thereby enabling saving used_aliquots in well directly through used_aliquot_attributes


app/models/pacbio/well.rb
```diff
@@ -111,7 +85,7 @@ def summary
 
     # A collection of all the used_aliquots for given libraries and pools in a well
     def base_used_aliquots
-      used_aliquots.collect(&:source).collect(&:used_aliquots).flatten
+      used_aliquots.collect(&:source).collect(&:used_aliquots).flatten.uniq(&:id)
```
BenTopping, line 88: This is a bit of a code smell. They should always be unique.

app/resources/v1/pacbio/run_resource.rb
```diff
@@ -63,7 +63,7 @@ def publish_messages
 
       private
 
-      def plates_attributes=(plates_parameters)
+      def plates_attributes=(plates_parameters) # rubocop:disable Metrics/MethodLength
```
BenTopping, line 66: Is this method length disable still needed?

spec/validators/well_validator_spec.rb
```diff
@@ -9,22 +9,34 @@
     context 'invalid' do
       it 'returns an error when there are multiple libraries but no tags' do
         # Going through the library_ids= method to create the aliquots
```
BenTopping, line 11: can remove comment

spec/validators/well_validator_spec.rb
```diff
@@ -9,22 +9,34 @@
...
-        well.library_ids = create_list(:pacbio_library, 2, tag:).collect(&:id)
+        libraries = create_list(:pacbio_library, 2, :tagged, tag:)
+        pool = create(:pacbio_pool, :tagged)
+        well.used_aliquots_attributes = [{ source_id: libraries[0].id, source_type: 'Pacbio::Library', volume: 10, concentration: 20, aliquot_type: :derived, template_prep_kit_box_barcode: '033000000000000000000' },
+                                         { source_id: libraries[1].id, source_type: 'Pacbio::Library', volume: 10, concentration: 20, aliquot_type: :derived, template_prep_kit_box_barcode: '033000000000000000000' },
+                                         { source_id: pool.id, source_type: 'Pacbio::Pool', volume: 10, concentration: 20, aliquot_type: :derived, template_prep_kit_box_barcode: '033000000000000000000' }]
+
         well.valid?
         expect(well.errors[:tags]).to include("are not unique within the libraries for well #{well.position}")
       end
```
BenTopping, line 42: library_ids and pool_ids in the below functions should be replaced by used_aliquots

spec/validators/well_validator_spec.rb
```diff
@@ -9,22 +9,34 @@
     context 'invalid' do
       it 'returns an error when there are multiple libraries but no tags' do
         # Going through the library_ids= method to create the aliquots
-        well.library_ids = create_list(:pacbio_library, 2, :untagged).collect(&:id)
+        library1 = create(:pacbio_library, :untagged)
+        library2 = create(:pacbio_library, :untagged)
+
+        well.used_aliquots_attributes = [{ source_id: library1.id, source_type: 'Pacbio::Library', volume: 10, concentration: 20, aliquot_type: :derived, template_prep_kit_box_barcode: '033000000000000000000' },
```
BenTopping, line 15: If you want to for these tests you could do `well.used_aliquots = build_list(:aliquot )...` instead of assigning to `used_aliquots_attributes`. The reason we used pool_ids and library_ids before was because it kept aliquots and libraries/pools in sync but that is no longer an issue.

spec/models/pacbio/run_spec.rb
```diff
@@ -242,7 +242,9 @@
     let!(:pools) { create_list(:pacbio_pool, 2) }
 
     it 'creates a run' do
```
BenTopping, line 244: pool_ids is also referenced in a test below and needs removal / replacement.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1357,6 +1377,19 @@
           expect(run.wells[1].column).to eq '12'
         end
 
+        it 'updates used_aliquots of well1 correctly' do
```
BenTopping, line 1380: Might be worth checking the number of used_aliquots is correct as well.

BenTopping (changes requested):
Looks really good, few minor comments and potentially an issue with validation as we discussed.

app/models/pacbio/well.rb
seenanair: Yes, this actually goes along with the issue we talked about where the validation is happening before the update. What it appears to me while debugging is that when we update an aliquot in a well using used_aliquots_attributes, for example, a well with two aliquots and we update one of those, it validates the well and it has three aliquots (two of them are the same, where one is the original object and one is the object to be updated).

app/resources/v1/pacbio/run_resource.rb
seenanair: It seems so, Rubocop is complaining about "Metrics/MethodLength: Method has too many lines."

app/models/pacbio/well.rb
```diff
@@ -51,39 +51,13 @@ class Well < ApplicationRecord
...
-      end
+    delegate :run, to: :plate, allow_nil: true
 
-      # If the used_aliquot is not in the list of ids, remove it
-      destroy_aliquots_by_source_type_and_id(ids, 'Pacbio::Library')
-    end
+    accepts_nested_attributes_for :used_aliquots, allow_destroy: true
 
     # Destroy aliquots based on their source_id and type
     # Used to keep libraries, pools and aliqouts in sync
```
BenTopping, line 63: This function can be removed.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1326,6 +1347,7 @@
 
         it 'has a ok status' do
           patch v1_pacbio_run_path(run), params: body, headers: json_api_headers
+          # byebug
```
BenTopping, line 1350: remove comment

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1382,12 +1402,18 @@
                   id: plate.id,
                   wells_attributes: [
                     {
-                      id: plate.wells.first.id.to_s,
-                      pool_ids: [plate.wells.first.pools[0].id, pool1.id]
+                      id: well1.id.to_s,
```
BenTopping, line 1405: I think we also want to check adding an aliquot with the same tag as a _destroy aliquot to test the validation doesn't break.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1411,19 +1437,20 @@
         it 'creates the correct number of used_aliquots' do
           expect do
             patch v1_pacbio_run_path(run), params: body, headers: json_api_headers
-          end.to change(Aliquot, :count).by(-7)
+          end.not_to change(Aliquot, :count)
```
BenTopping, line 1440: If we are removing aliquots shouldn't this be changed? Or are we adding the same amount we are removing?

BenTopping (changes requested):

spec/requests/v1/pacbio/runs_spec.rb
seenanair: Yes, we are adding 1 and removing 1.

app/models/pacbio/well.rb
```diff
@@ -111,7 +77,7 @@ def summary
 
     # A collection of all the used_aliquots for given libraries and pools in a well
     def base_used_aliquots
-      used_aliquots.collect(&:source).collect(&:used_aliquots).flatten
+      used_aliquots.reject(&:marked_for_destruction?).collect(&:source).collect(&:used_aliquots).flatten # rubocop:disable Layout/LineLength
```
BenTopping, line 80: you can have this span multiple lines and remove the rubocop line layout. If you remove the comment and run `bundle exec rubocop -A` that will attempt to autofix errors like this.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -758,14 +761,21 @@
...
+                                                 { source_id: pool2.id, source_type: 'Pacbio::Pool', volume: 10, concentration: 20, aliquot_type: :derived, template_prep_kit_box_barcode: '033000000000000000000' }] }
                   ]
                 }]
               }
             }
           }.to_json
         end
 
+        before do
+          if pool2.used_aliquots.any? && pool1.used_aliquots.any?
```
BenTopping, line 774: Is this `if` needed? Conditional checks for setting up test data can lead to flakey tests. If this is needed for a specific test then include this in that test.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -828,14 +839,21 @@
...
+                                                 { source_id: library2.id, source_type: 'Pacbio::Library', volume: 10, concentration: 20, aliquot_type: :derived, template_prep_kit_box_barcode: '033000000000000000000' }] }
                   ]
                 }]
               }
             }
           }.to_json
         end
 
+        before do
+          if library1.used_aliquots.any? && library2.used_aliquots.any?
```
BenTopping, line 852: Same comment as above

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1411,19 +1437,83 @@
...
+            data: {
+              id: run.id,
+              type: 'runs',
+              attributes: {
+                plates_attributes: [{
+                  id: plate.id,
+                  wells_attributes: [
+                    {
+                      id: well1.id.to_s,
+                      used_aliquots_attributes: well1.used_aliquots.map(&:attributes) + [aliquot_to_update]
```
BenTopping, line 1484: should this aliquot to update be in well2 since its tag is the same as a well2 aliquot and well2 is the one with destruction?

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1411,19 +1437,83 @@
...
+          patch v1_pacbio_run_path(run), params: body, headers: json_api_headers
+          run.reload
+          updated_plate = run.plates.find_by(id: plate.id)
+          expect(updated_plate.wells.length).to eq 1
+          expect(updated_plate.wells[0].pools.length).to eq 2
+          expect(updated_plate.wells.first.used_aliquots.length).to eq(well1.used_aliquots.length + 1)
+          updated_well_aliquot = updated_plate.wells.first.used_aliquots.last
+          expect(updated_well_aliquot.tag_id).to eq well2.used_aliquots[0].tag_id
+          aliquot_to_update.each do |key, value|
+            next if key == :aliquot_type
```
BenTopping, line 1513: why do we need to skip the aliquot_type

BenTopping (changes requested):
Looking good, just a couple questions around the test setup.

app/models/pacbio/well.rb
seenanair: I am still getting that error even after running bundle exec rubocop -A

spec/requests/v1/pacbio/runs_spec.rb
seenanair: No, this one is creating a new array that contains all the existing aliquots already in well1 plus the new aliquot  to update (aliquot_to_update) which contains the tag which is same as well2.
I hope this makes sense

spec/requests/v1/pacbio/runs_spec.rb
seenanair: For aliquot_type we need to pass in ':derived" but the returned value is 'derived'. So, it is easier to match all other updated values in one go, if we skip it

spec/requests/v1/pacbio/runs_spec.rb
BenTopping: I may be wrong but I think you can have the same tag across multiple wells its just you cant have the same tag twice in the same well.

seenanair:
@BenTopping One more question: Do we need an end-to-end test to check the updating of 'available volume,' especially to verify the scenario where the same library is used in multiple wells, and then we add or remove them to ensure the available volume is correct in context?

seenanair:
@stevieing Please have a look into this PR when you get time.

stevieing (approved):

PR #1329, opened 2024-05-29
DPL-671 - Prevent hidden tag sets from being used

### LINKED ISSUE #920: DPL-671 - As a developer I need to be able to deactivate tag groups so that tag groups that are no longer used can be hidden
**User story**
As a developer I need to be able to deactivate tag groups so that tag groups that are no longer used can be hidden so users will only see tag sets that are currently used.

If a tag set has previously been used it cannot be deleted.

Steve

Steve

**Acceptance criteria**
- [ ] The tag set should have an active tag which can be true or false
- [ ] Any existing API end points should only return active tag sets

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

SHIV5T3R:
Closes #920 

- Tag sets now have an `active` attribute that defaults to true, along with a method to set the value
- Only active tag sets are returned by the API

### DIFF
--- a/app/models/tag_set.rb
+++ b/app/models/tag_set.rb
@@ -9,5 +9,12 @@ class TagSet < ApplicationRecord
   enum sample_sheet_behaviour: { default: 0, hidden: 1 }, _suffix: true
 
   validates :name, presence: true
-  validates :pipeline, presence: true
+  
+  def set_active
+    update(active: true)
+  end
+  
+  def set_inactive
+    update(active: false)
+  end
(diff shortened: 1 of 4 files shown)


SHIV5T3R:
One thing I wanted to ask about is that rubocop made me change the setter method to something that isn't `set_attribute`, what's the standard in Ruby or across projects here?

PR #1340, opened 2024-06-11
1179 dpl 1041 as a stakeholder james w i would traction to track pacbio smrtlink library types

stevieing:
Closes #

- Add library type to smrt link options
- Add rake task to update library types
- Move smrt link options for well store to configuration


lib/tasks/update_smrt_link_option_library_type_for_revio.rake
BenTopping, line 1: Wondering if we can add this as part of the create_smrt_link_versions rake task so all existing records matching the smrt_link_version get the default when a new one is added? Would make future changes simpler as well.

app/models/pacbio/well.rb
```diff
@@ -23,23 +23,10 @@ class Well < ApplicationRecord
...
-                        binding_kit_box_barcode
-                        pre_extension_time
-                        loading_target_p1_plus_p2
-                        movie_time
-                        movie_acquisition_time
-                        include_base_kinetics
-                        library_concentration
-                        polymerase_kit]
+    # We now get the accessors from configuration
+    store :smrt_link_options
```
BenTopping, line 27: Nice.

lib/tasks/update_smrt_link_option_library_type_for_revio.rake
stevieing: Not entirely sure what you mean. If we run that task again I think it would just add the new option as it is already in there?

BenTopping: Yes it adds the new option but I was wondering whether the data should also be seeded there as well. So when we add a new option and then run `create_smrt_link_versions rake task` it would also set the default in the appropriate wells similar to how this rake task is currently doing it.
Saves us needing to do a separate seed rake task for the data whenever we add a new option.

stevieing: Ok. That makes sense.

BenTopping, line 1: Can this be removed now?

BenTopping (changes requested):

lib/tasks/update_smrt_link_option_library_type_for_revio.rake
stevieing: Yes. I had removed the tests which is why codecov is failing.

BenTopping (approved):

BenTopping (approved):

BenTopping (approved):

PR #1341, opened 2024-06-11
Y24-126-1 - Support automatic pool creation in reception

### LINKED ISSUE #1339: Y24-126-1  - As a developer (Ben) I would like to support automatic pool creation for ONT library tubes imported from Sequencescape.
**User story**
As a developer (Ben) I would like to support automatic pool creation for ONT library tubes imported from Sequencescape.

Danni W, Tavis M, Ben T

Danni W, Tavis M

**Acceptance criteria**
- [ ] Add an option to the reception api to allow pool creation.
- [ ] Reception pool creation should create an ONT pool with all the samples of that reception import.
  - First create all the tubes, samples and requests. Using the imported data create a pool; the pool barcode should be the same as the Sequencescape library tube barcode.
  - Find ONT tags based on the imported sample sequence (will be included in the reception message). Prior to this work check there are no duplicate sequences as otherwise we may need to add extra steps like tag selection in the UI.
  - ONT Pool information (conc, volume, insert_size and kit barcode (tag set name)) can be pulled from one of the samples as they should all be the same.
- [ ] Ensure there is suitable error handling e.g. tag_set could not be found from sample...

References
This story is the product of the research story [DPL-1018](https://github.com/sanger/traction-ui/issues/1434)

BenTopping:
Closes #1339

- Adds pool_attributes to reception api
- Enables ONT pool tube barcode to be overridden
- Enables ONT pools to be created via the reception

This PR's code can be reviewed but it is reliant on [Y24-126-2](https://github.com/sanger/traction-ui/issues/1747) being completed for manual end to end testing.

### DIFF
--- a/app/ont/ont.rb
+++ b/app/ont/ont.rb
@@ -31,9 +31,11 @@ module Ont
   end
 
   def self.library_factory(request:, library_attributes:)
-    # We need to find the tag_id from the tag_sequence provided
-    # TODO: We should take a tag_set here as well as there are duplicate oligos across tagsets
-    library_attributes[:tag_id] = Tag.find_by(oligo: library_attributes[:tag_sequence]).id
+    # Get the tag_set from the kit_barcode
+    ont_tag_set = TagSet.find_by(name: library_attributes[:kit_barcode])
+    # Find the tag_id from the tag_set based on the tag_sequence/oligo
+    library_attributes[:tag_id] =
+      ont_tag_set.tags.find_by(oligo: library_attributes[:tag_sequence])&.id
     filtered_attributes = library_attributes.slice(*self.library_attributes)
 
     Ont::Library.new(
--- a/app/ont/ont.rb
+++ b/app/ont/ont.rb
@@ -7,7 +7,9 @@ module Ont
   end
 
   def self.library_attributes
-    raise StandardError, 'Unsupported' # Only Pacbio is supported at the moment
+    %i[
+      volume concenetration kit_barcode insert_size tag_id
+    ]
   end
 
   def self.request_attributes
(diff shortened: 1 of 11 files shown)


spec/lib/tasks/volume_tracking.rake_spec.rb
```diff
@@ -22,7 +24,7 @@
       well_aliquots = create_list(:aliquot, 10, used_by: create(:pacbio_well), volume: 10, concentration: 10)
 
       # We shouldnt change the amount of aliquots
-      expect { Rake::Task['volume_tracking:clear_well_aliquot_volume'].invoke }.not_to change(Aliquot, :count) and output(
+      expect { Rake::Task['volume_tracking:clear_well_aliquot_volume'].invoke }.to not_change(Aliquot, :count).and output(
```
seenanair, line 27: Why is this change necessary if they both achieve the same thing, I hope?

seenanair (commented):
Looks fine to me, but I don't have enough Ruby/Rails expertise to verify it thoroughly

spec/lib/tasks/volume_tracking.rake_spec.rb
BenTopping: Yeah they are the same thing, this is a slightly unrelated change but this test was outputting text when running the test suite which we don't want and this fixes it.

app/ont/ont.rb
```diff
@@ -24,9 +30,18 @@ def self.direct_request_attributes
...
-  # Parameters would be request: and library_attributes: if this was supported.
-  def self.library_factory(*)
-    raise StandardError, 'Unsupported' # Only Pacbio is supported at the moment
+  def self.library_factory(request:, library_attributes:)
+    # Get the tag_set from the kit_barcode
+    ont_tag_set = TagSet.find_by(name: library_attributes[:kit_barcode])
+    # Find the tag_id from the tag_set based on the tag_sequence/oligo
+    library_attributes[:tag_id] =
+      ont_tag_set&.tags&.find_by(oligo: library_attributes[:tag_sequence])&.id
+    filtered_attributes = library_attributes.slice(*self.library_attributes)
```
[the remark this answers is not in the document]
BenTopping: In theory tag id can be nil if they are importing just one library in a pool so its a bit of a tricky situation. If there are multiple libraries and some dont have tags it will get caught later on in validation and error.

seenanair (approved):

Issue #1322, opened 2024-05-20
Y24-096: Publish volume tracking message from Traction Service


dasunpubudumal:
**User story**
As a developer, I need to setup a connection between Traction Service and the Enterprise Message Queue (EMQ) so that I can publish volume tracking messages into EMQ which are then consumed by ToL Lab Share. 

I need to use the exchange and queues that are already set up in https://github.com/sanger/General-Backlog-Items/issues/397 to set up a connection, and publish the aliquot creation message discussed in https://github.com/sanger/tol-lab-share/issues/57 to EMQ.

The schema for the message is defined in the story https://github.com/sanger/General-Backlog-Items/issues/396, and needs to be validated by fetching the schema from RedPanda schema store.

Note that stories https://github.com/sanger/traction-service/issues/1202 and https://github.com/sanger/traction-service/issues/1206 are focused on sending the messages from Traction. Therefore, the scope of this particular story is only confined to establishing the connection and sending a rough message into the EMQ. Within the scope of this story, I do not have to consider _when_ I need to send a volume tracking message. That will be implemented in  https://github.com/sanger/traction-service/issues/1202 and https://github.com/sanger/traction-service/issues/1206.

**Who are the primary contacts for this story**
Dasun, Stuart, Steve

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] Create a user account for traction in deployment project for EMQ
- [ ] To establish a connection between Traction Service and EMQ to publish messages into the EMQ.
- [ ] To publish an aliquot creation message using the Traction Service, you should design a generic method for creating messages using existing config files and then publish them
 This could be done using the Rails console, until stories https://github.com/sanger/traction-service/issues/1202 and https://github.com/sanger/traction-service/issues/1206 are done. 
**Dependencies**
This story is blocked by the following dependencies:
- https://github.com/sanger/General-Backlog-Items/issues/396
- https://github.com/sanger/General-Backlog-Items/issues/397

References
This story has a non-blocking relationship with:
- https://github.com/sanger/tol-lab-share/issues/57
- https://github.com/sanger/traction-service/issues/1202
- https://github.com/sanger/traction-service/issues/1206

**Additional context**

Diagrams: 

[...]

PR #1358, opened 2024-07-12
Y24-096 publish volume tracking message

seenanair:
Closes [#1322](https://github.com/sanger/traction-service/issues/1322)

Following are the changes proposed in this pull request
- Create a config based aliquot creation message for volume tracking using the existing framework, which aligns with the Avro schema definition (re-used code from SS implementation of similar work by Stuart)
- Update  corresponding model classes (aliquot, library, pool, well) with methods to generate required field vales for the message
- Validate the message using Avro schema defined in Redpanda registry (re-used code from SS implementation of similar work by Stuart)
- Create a class responsible for establishing connection and publishing messages to EMQ
- Create a class responsible for Avro validation
- Define a public Interface to allow publishing message from resource
- Update bunny.yml and bunny.example.yml with RabbitMQ and Redpanda configuration settings


app/messages/emq/publishing_job.rb
```diff
@@ -0,0 +1,61 @@
...
+        sender.send_message(message)
+      end
+    end
+
+    # recursively converts a nested hash into an OpenStruct,
+    # allowing for dot notation access to hash keys and their values.
+    def self.deep_open_struct(obj)
+      return obj unless obj.is_a?(Hash)
+
+      OpenStruct.new(obj.transform_values { |val| deep_open_struct(val) }) # rubocop:disable Style/OpenStructUse
```
dasunpubudumal, line 58: Is there a special need for `OpenStruct` instead of a normal `Hash` here? I've heard that these can add some kinds of [performance bottlenecks](https://ruby-doc.org/stdlib-1.9.3/libdoc/ostruct/rdoc/OpenStruct.html#class-OpenStruct-label-Implementation-3A). 

Note: It won't matter much because we're not dealing with a large corpus of data at this stage.

app/messages/emq/publishing_job.rb
```diff
@@ -0,0 +1,61 @@
...
+      # Check if the schema_key exists in the subjects hash and return early if it does not
+      schema = bunny_config.amqp.schemas.subjects[schema_key]
+      return if schema.nil?
+
+      # Get the subject and version from the schema and return early if either is nil
+      subject = bunny_config.amqp.schemas.subjects[schema_key].subject
+      version = bunny_config.amqp.schemas.subjects[schema_key].version
+      return if subject.nil? || version.nil?
+
+      # Create a validator and sender for the subject and version
```
dasunpubudumal, line 33: Looks like the `Emq::Validator` class validates the message, doesn't it? `subject` and `version` are validated before in line 31? Do you think it's better to be renamed to something like `Emq::MessageValidator`?

app/messages/emq/publishing_job.rb
```diff
@@ -0,0 +1,61 @@
...
+        message_object = Message::Message.new(object:,
+                                              configuration: create_messsage_configuration).content
+
+        # check if the schema_key is present in the payload
+        next if message_object[schema_key].nil?
+
+        # Validate the message against the schema and send it to the EMQ
+        publish_message = message_object[schema_key]
+        message = validator.validate_message(publish_message)
+        sender.send_message(message)
```
dasunpubudumal, line 49: One thing that we might have to consider is what are we going to do in case this invocation fails (e.g. due to RMQ unavailability, some networking issue)? 

I think it needs to be logged properly? Saying clearly that this message with a certain identifier was failed or something?

app/messages/emq/validator.rb
```diff
@@ -0,0 +1,92 @@
...
+      encoder.write([schema.crc_64_avro_fingerprint].pack('Q')) # 8 byte schema fingerprint
+      writer.write(message, encoder)
+      stream.string
+    end
+
+    # Create the message schema
+    # @return [String] the schema for the message
+    def create_message_schema
+      # Prefer to use the cached schema if it exists.
+      cache_file_path = "data/avro_schema_cache/#{@subject}_v#{@version}.avsc"
```
dasunpubudumal, line 58: Using a cached schema is really nice. Less networking overhead :)

app/messages/emq/validator.rb
```diff
@@ -0,0 +1,92 @@
...
+      if File.exist?(cache_file_path)
+        Rails.logger.debug { "Using cached schema for #{@subject} v#{@version}" }
+        return File.read(cache_file_path)
+      end
+
+      # Default to fetching the schema from the registry and caching it.
+      Rails.logger.debug { "Fetching and caching schema for #{@subject} v#{@version}" }
+      response = fetch_response("#{@registry_url}#{@subject}/versions/#{@version}")
```
dasunpubudumal, line 66: If the response err, do we need to handle it separately? For example, if the `fetch_response` function returns at line 88, does `response.body` return `nil`? Not sure how `JSON.parse()` behave under `nil` inputs.

dasunpubudumal (commented):
The implementation is really nice. I added some comments, but most of them are clarifications for myself :D

app/models/aliquot.rb
```diff
@@ -59,4 +59,43 @@ def tagged?
   def collection?
     false
   end
+
+  # Returns the id to publish
```
stevieing, line 63: It would feel better to have all the published data together in one object as it would all be called together.

app/models/aliquot.rb
```diff
@@ -59,4 +59,43 @@ def tagged?
...
+  def publish_id
+    id.to_s
+  end
+
+  # Returns the type of the source to publish
```
stevieing, line 68: Same here and below. Also why are using respond_to here? Is it necessary. It feels better to have a publish or publisher object. Also it might be better to do the to_s or '' in the source.

app/models/aliquot.rb
```diff
@@ -59,4 +59,43 @@ def tagged?
...
+  # Returns the sample name of the source to publish
+  def publish_sample_name
+    if source.respond_to?(:publish_data_source)
+      source.publish_data_source[:sampleName].to_s
+    else
+      ''
+    end
+  end
+
+  # Returns the barcode of the used_by to publish
```
stevieing, line 87: Same again. Not sure why we need the respond to and it feels like anything to do with publishing should be in an object or nested class.

app/models/pacbio/library.rb
```diff
@@ -93,6 +93,23 @@ def sequencing_plates
       wells&.collect(&:plate)
     end
 
+    # Return the data to be published as the source of an aliquot
```
stevieing, line 96: Same again. It does feel like publish_data should be a single object / class / hash.

app/models/pacbio/pool.rb
```diff
@@ -66,6 +66,14 @@ def collection?
       used_aliquots.length > 1
     end
 
+    # Return the data to be published as the used_by of an aliquot
```
stevieing, line 69: Same as before a wrapper would be better.

app/models/pacbio/well.rb
```diff
@@ -140,5 +140,13 @@ def update_smrt_link_options(options)
       end
       save!
     end
+
+    # Return the data to be published as the used_by of an aliquot
```
stevieing, line 144: Same as abov.e

config/pipelines/pacbio.yml
```diff
@@ -298,8 +298,54 @@ default: &default
 
       # No changes required from v12_sequel_iie, added for consistency
       v13_sequel_iie: *v12_sequel_iie
-
-
+      
+    # For any EMQ message creation, the section and key name must be same as schema subject in bunny.yml
+    volume_tracking:
```
stevieing, line 303: Nice.

stevieing (changes requested):
Mostly stuff around how we implement the publisher methods and classes.

app/messages/emq/publishing_job.rb
seenanair: I used OpenStruct because 1) it was already in use and 2)it provides a more cleaner syntax and convenience, especially with dot notation access as the configuration data structure id nested. I read about the performance issue, thought it wouldn't be an issue as the structure is small.
What do you think?

app/messages/emq/publishing_job.rb
seenanair: Sure, will do 👍

app/messages/emq/publishing_job.rb
seenanair: Ok, I will see, how it can be changed.

app/messages/emq/validator.rb
seenanair: Credit goes to @sdjmchattie as I used it in the same way as he did in Sequencescape. :)

app/messages/emq/publishing_job.rb
dasunpubudumal: Yeah it doesn't really matter much at this stage. We can keep it the way it is.

app/messages/emq/publishing_job.rb
sdjmchattie: I can see why you've suggested this because it may look like you're validating the subject and version.  I wonder whether a better solution is to not pass in the subject, version and schema registry and instead have a different module or class deal with getting the schema.  Then pass the schema to the validator's constructor.

app/messages/emq/publishing_job.rb
```diff
@@ -0,0 +1,61 @@
...
+        # Construct the message to publish from the object using the given configuration
+        message_object = Message::Message.new(object:,
+                                              configuration: create_messsage_configuration).content
+
+        # check if the schema_key is present in the payload
+        next if message_object[schema_key].nil?
+
+        # Validate the message against the schema and send it to the EMQ
+        publish_message = message_object[schema_key]
+        message = validator.validate_message(publish_message)
```
sdjmchattie, line 48: I'm not sure I would have called this a validator because you're clearly using it to actually encode the message here.  I would have called this the encoder.  Part of encoding is validating that the message can be encoded which is what you're calling validation.

app/models/aliquot.rb
seenanair: Totally agree, and I really wanted to do it that way and started of  with a 'publishing_data ' method in aliquot. I tried it, but didn't work  with how the DatastructureBuilder is written. To my understanding this requires changes in the current DataStructureBuilder class where a '.' access with a 'model' type recursively send it to the object itself.

app/models/pacbio/library.rb
```diff
@@ -93,6 +93,23 @@ def sequencing_plates
...
+    def publish_data_source
+      {
+        sourceType: 'library',
+        sourceBarcode: tube&.barcode,
+        sampleName: request.sample_name
+      }
```
sdjmchattie, line 102: My concern with doing this here is that you're tying the Traction Service models to one particular Avro message schema.  We made this mistake in the TOL-lab-share system when it pushes messages out to Traction.  Ideally I think you should have a separate class for turning your model into the specific Avro schema format.

If that could be something config driven, a schema for how to map values, that would be ideal, but might be a bit much for a first attempt at this.

sdjmchattie (changes requested):

app/models/pacbio/library.rb
stevieing: I think it might be worth going through this to see what we can do.

app/messages/emq/publishing_job.rb
sdjmchattie: (Or serializer works as well.  I've often thought of it as encoding though.)

app/models/pacbio/library.rb
seenanair: @sdjmchattie  @stevieing I can see the point where Avro file format has significantly influenced  model objects.
I started with an approach similar to yours, where  a class to generates a message that conforms to the Avro format, but not using config. After introducing config, it was handed over the responsibility to model objects as which was the main purpose or effect of using config. 
So, you are suggesting  merging these two ideas?  Specifically, using a config driven approach where config delegates to a specific class responsible for generating this message. Is tat correct?

sdjmchattie: I worked on a similar mapping problem about 7 years ago where it was in C# and we used something called AutoMapper.  I don't remember the specific, but I have some recollection of having a class that deals exclusively with one message type you want to create, and you were able to define what it should do via a config something a bit like:

```
library.tube&.barcode => sourceBarcode
library.request.sample_name => sampleName
```

I'm not sure whether there's any gem we can use in Ruby to support such a mapping operation.  What I've shown above is very much an oversimplification as I know there are arrays and other transformations that values might need to have done to them, so I don't think we should roll our own.  However, even if you just had a message generator class that you pass it the library, aliquot, whatever else is needed to fully form the message.  It would be OK for that class to know how to dissect the model and get the data needed.

It would be worthwhile commenting in the models that this class exists so any changes should be cross referenced to make sure the class is able to continue doing its job.  That would probably be caught by good unit tests though.

Another considering could be to have extensions for this class that make it mappable to the desired Avro format.  That way you keep the main model file clean but `include` the extensions to gain access to internal data.  Not sure if this is just adding more complexity though.

app/models/aliquot.rb
seenanair: I think this requires changes in DataStructure builder, similar to one I mentioned above. That's is the reason why I have made it work in such a way to avoid changes in Datastructure builder as this story has already gone beyond the size. But with the new suggestions, I think that is required any way, so let me see how these can be resolved.

app/models/pacbio/library.rb
seenanair: Ok, There is already a class responsible for the mapping operation that we created. It may be best to extend this class so that it can handle the specific job of delegating from where these model methods are accessed. This approach would likely be the most viable starting point, I think.

seenanair: The refactoring includes a config-based approach to create the message without modifying model objects. This approach uses the yaml configuration to dynamically create an object responsible for message building, using a class specified in the configuration. It delegates the value creation of fields specified in the config to this object.  
The Avro schema version is included in the configuration to ensure that the schema it is created against is the same one configured in bunny. 
I hope that anything in the future that uses the 'EMQ::Publisher' can follow a similar structure, with new sections added against schema keys and versions in yml file.

app/messages/emq/publishing_job.rb
seenanair: My fault:  I was focused on validation rather than the real job all the time. Renamed to encoder.

app/messages/emq/validator.rb
seenanair: Checked this and it will be thrown as part of Standard Error handler (which already exists) it seems.

seenanair:
> Mostly stuff around how we implement the publisher methods and classes.

@stevieing Removed those as part of the new refactoring.

sdjmchattie (approved):
One final recommendation, but otherwise looks much better.  Thanks for taking the time to make some non-trivial changes!

seenanair:
> One final recommendation, but otherwise looks much better. Thanks for taking the time to make some non-trivial changes!

Thank you for the feedback. Happy that it is looking better now.

stevieing (approved):
Really good.

We could probably do with looking at some of the Rubocop issues but maybe worth delaying until we touch his again.

You have done more than enough for this story!

dasunpubudumal (approved):

dasunpubudumal (approved):

PR #1363, opened 2024-07-22
feat: Update used aliquots data when primary aliquot is updated for a library

stevieing:
Closes #

- 

### DIFF
--- a/app/models/pacbio/library.rb
+++ b/app/models/pacbio/library.rb
@@ -49,6 +49,8 @@ module Pacbio
     # if the new volume is greater than the used volume.
     before_update :primary_aliquot_volume_sufficient
 
+    after_update :update_used_aliquots
+
     before_destroy :check_for_derived_aliquots, prepend: true
 
     def create_used_aliquot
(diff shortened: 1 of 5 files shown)


app/models/pacbio/library.rb
```diff
@@ -49,6 +49,8 @@ class Library < ApplicationRecord
     # if the new volume is greater than the used volume.
     before_update :primary_aliquot_volume_sufficient
 
+    after_update :update_used_aliquots
```
BenTopping, line 52: Will this run before this https://github.com/sanger/traction-service/blob/develop/app/resources/v1/pacbio/library_resource.rb#L39? Could we add a test for it as this resource method should update the warehouse with the updated tag data.

BenTopping (commented):
Looks good just one comment about potential race condition.
Do we also need a script to fix the prod data?

app/models/pacbio/library.rb
stevieing: Not sure if this is easily testable but "in Ruby on Rails, callbacks such as after_update run within the transaction of the update action. This means if you have an after_update callback that makes changes to the database and an error occurs after the update action but within the same transaction, all changes including those made in the after_update callback will be rolled back. This ensures data integrity by not applying partial updates to the database."

We can test on UAT and training.

BenTopping: Sure, yeah easiest way will be to try on UAT. You might be able to manually test it by putting a debugger in the resource after_update method and checking if the data there has been updated.

BenTopping (approved):

PR #1365, opened 2024-07-23
refactor: add uuid field to aliquot

seenanair:
Closes https://github.com/sanger/traction-service/issues/1322

- Added uuid field to aliquot
- Changed Avro schema message field aliquotId to aliquotUuid


stevieing (changes requested):
All good but we will need a rake task to update all existing aliquots so the messages will work.

dasunpubudumal (approved):

lib/tasks/update_aliquots.rake
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+namespace :update_aliquots do
```
sdjmchattie, line 3: Personally I would call this `aliquots` and then call the task `add_missing_uuids`.  That makes it very clear when this gets called like `bundle exec rake aliquots:add_missing_uuids`

This is consistent with the way rails uses tasks like `db:setup` and `assets:clobber`.  You have a noun on the left and a verb on the right.

lib/tasks/volume_tracking.rake
```diff
@@ -16,4 +16,15 @@ namespace :volume_tracking do
       end
     end
   end
+
+  task update_aliquots_with_uuid: :environment do
```
sdjmchattie, line 20: `add_missing_uuids_to_aliquots`.  Would this be better calling your other rake task instead of adding the implementation a second time?

See https://stackoverflow.com/a/1290119/772095 showing you should just be able to add it similarly to the way `:environment` is specified here.  You can also do it explicitly with `Rake::Task['task_name'].invoke` inside the block.

sdjmchattie (changes requested):

lib/tasks/update_aliquots.rake
seenanair: Agree, it makes it more clear 👍

lib/tasks/volume_tracking.rake
seenanair: Good spot. Sorry, I think this is not required here. That was something added initially which was not removed.

sdjmchattie (approved):

stevieing (approved):

PR #1372, opened 2024-07-26
Sending retention instructions to the warehouse

### LINKED ISSUE #1349: Y24-161 - Update sending the retention instructions with the create labware message to mlwh
**User story**
As developers, we need to send the retention instruction to mlwh in the `create-labware` message for each labware created when samples are received through the ToL automated manifest messages to allow for correct audit tracking. Use the changes done in the schema in https://github.com/sanger/General-Backlog-Items/issues/412 story.

Steve / Stuart / Dasun

**Acceptance criteria**
- [x] The retention instructions should be one of the following values listed in https://github.com/sanger/tol-lab-share/issues/85#issuecomment-2107445987.

**Dependencies**
- https://github.com/sanger/General-Backlog-Items/issues/412
- https://github.com/sanger/tol-lab-share/issues/121

References
- https://github.com/sanger/tol-lab-share/issues/85

**Additional context**

dasunpubudumal:
Closes #1349 

- Adding a migration to store retention instructions in a `retention_instruction` column in `samples` table.
- Model adding retention instructions as done in sequencescape
- Update the messaging mechanism to include the `retention_instruction` to be sent to the warehouse queue.


sdjmchattie (commented):
I don't see any changes in here that tell Traction Service to use version 2 of the schema?

sdjmchattie (approved):

spec/models/sample_spec.rb
```diff
@@ -59,4 +65,24 @@
       expect(sample.requests.length).to eq 2
     end
   end
+
+  context 'after create' do
```
BenTopping, line 69: Does this test anything differently to the added test above? Not sure whether its needed.

spec/models/sample_spec.rb
```diff
@@ -43,6 +43,12 @@
         expect(build(:sample, species: nil)).not_to be_valid
       end
     end
+
+    describe 'retention_instruction' do
+      it 'has a retention_instruction' do
```
BenTopping, line 48: What happens if the retention instruction received is invalid?

BenTopping (approved):
Looks good.

app/resources/v1/reception_resource.rb
```diff
@@ -77,7 +77,7 @@ def permitted_request_attributes
     def permitted_sample_attributes
       %i[name external_id species study_uuid priority_level public_name
          sanger_sample_id supplier_name taxon_id donor_id country_of_origin
-         accession_number date_of_sample_collection]
+         accession_number date_of_sample_collection retention_instruction]
```
seenanair, line 80: Just a doubt - I hope we don't need a separate validation for retention_instruction here as it will be handled as part of the schema validation for any unexpected values.

seenanair (approved):

app/resources/v1/reception_resource.rb
sdjmchattie: Avro isn't involved in the creation of the message being processed here, so it could very easily not pass validation.

spec/models/sample_spec.rb
dasunpubudumal: I wanted to have an assertion to test the cases before Rails type cast the enums. I will remove the first example (which is redundant).

app/resources/v1/reception_resource.rb
sdjmchattie: It is, but then you map the values around and so there's currently nothing stopping you submitting anything you want to this Traction endpoint.  Plus the implementation here doesn't know you put it through Avro.  If in future we change the way we create the messages coming into this API, we'd like to have confidence that we can't submit values that cause issues.

dasunpubudumal: That is quite true. I understand now what you meant by Avro isn't involved here. 

I'll probably add some validations for the retention instructions. I'm not sure if any other system calls this endpoint but in that case we'll need to have some validations for retention instructions here.

spec/models/sample_spec.rb
dasunpubudumal: I was under the impression that `/v1/reception` endpoint was used only by `tol-lab-share`. There are [strict validations](https://redpanda.uat.psd.sanger.ac.uk/console/) when this message is consumed, but not when it is pushed to traction. Although it is not pushing invalid values from `tol-lab-share` currently, there's nothing preventing any other developments in `tol-lab-share` (or any other component that pushes HTTP calls to `/v1/reception` endpoint from pushing an invalid retention instruction. 

I will add some validations here.

app/models/sample.rb
```diff
@@ -6,9 +6,22 @@ class Sample < ApplicationRecord
 
   attr_readonly :name
 
+  enum retention_instruction: {
+    destroy_after_2_years: 0,
+    return_to_customer_after_2_years: 1,
+    long_term_storage: 2
+  }
+
+  VALID_RETENTION_INSTRUCTIONS = %w[
```
sdjmchattie, line 15: I have a feeling you can just use `Sample.retention_instructions.keys`

app/models/sample.rb
```diff
@@ -6,9 +6,22 @@ class Sample < ApplicationRecord
...
+    destroy_after_2_years
+    return_to_customer_after_2_years
+    long_term_storage
+  ].freeze
+
   validates :species, presence: true
   validates :external_id, presence: true, uuid: true
   validates :name, uniqueness: { case_sensitive: false }, presence: true
+  validates :retention_instruction, inclusion: { in: VALID_RETENTION_INSTRUCTIONS }, allow_nil: true
```
sdjmchattie, line 24: You probably need some tests to go with this to see how it handles a message actually trying to set an invalid value.  I'm not entirely sure where this will break.  The whole point of an enum is that you cannot set invalid values so it may break before this validation even occurs.

spec/models/reception_spec.rb
```diff
@@ -37,6 +37,28 @@
...
+              request: { library_type: 'lib', data_type: 'dt', cost_code: generate(:cost_code), external_study_id: generate(:external_study_id) },
+              sample: { name: generate(:sample_name), species: 'Human', external_id: generate(:uuid), retention_instruction: 'invalid_instruction' }
+            }
+          ]
+        }
+      ], tubes_attributes: [] }
+    end
+
+    it 'raises an ArgumentError' do
+      expect { reception.valid? }.to raise_error(ArgumentError)
```
sdjmchattie, line 58: So I don't think this is the validation you wrote causing this, am I right?  If so, then you don't need to validate the value in that model attribute.

dasunpubudumal: I think what's causing the `ArgumentError` is the `enum` mapping defined in the model. The intention behind this test is to check the validity of a reception resource when an invalid `sample` is sent into the endpoint.

I didn't quite get what you meant by:

> ... then you don't need to validate the value in that model attribute.

sdjmchattie: What I'm saying is that if you removed the line `validates :retention_instruction, inclusion: { in: VALID_RETENTION_INSTRUCTIONS }, allow_nil: true` from the model, I suspect the tests will still show that `ArgumentError` occurs when you want it to.  I might be wrong.

dasunpubudumal: Nope you're correct. I just forgot to remove it. The `enum` mapping does the validation; it doesn't require an additional callback in this case.

sdjmchattie (approved):
Looks good

PR #1379, opened 2024-08-01
DPL-1074 sent library volume to warehouse while creating or updating a sequencing run

seenanair:
Closes https://github.com/sanger/traction-service/issues/1206

- Publish message to EMQ for all library aliquots when a run is created or updated

### DIFF
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -35,7 +35,8 @@ module VolumeTracking
       when 'Pacbio::Well'
         data[:used_by_type] = 'run'
         data[:used_by_barcode] =
-          "#{aliquot.used_by.plate.sequencing_kit_box_barcode}:#{aliquot.used_by.plate.plate_number}:#{aliquot.used_by.position}" # rubocop:disable Layout/LineLength
+          "#{aliquot.used_by.plate.sequencing_kit_box_barcode}:" \
+          "#{aliquot.used_by.plate.plate_number}:#{aliquot.used_by.position}"
       when 'Pacbio::Pool'
         data[:used_by_type] = 'pool'
         data[:used_by_barcode] = aliquot.used_by.tube.barcode
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -33,7 +33,7 @@ module VolumeTracking
 
       case aliquot.used_by_type
       when 'Pacbio::Well'
-        data[:used_by_type] = 'well'
+        data[:used_by_type] = 'run'
         data[:used_by_barcode] =
           "#{aliquot.used_by.plate.sequencing_kit_box_barcode}:#{aliquot.used_by.plate.plate_number}:#{aliquot.used_by.position}" # rubocop:disable Layout/LineLength
       when 'Pacbio::Pool'
--- a/app/resources/v1/pacbio/run_resource.rb
+++ b/app/resources/v1/pacbio/run_resource.rb
@@ -56,12 +56,7 @@ module V1
 
       def publish_messages
         Messages.publish(@model, Pipelines.pacbio.message)
-        all_library_aliquots = @model.plates.flat_map do |plate|
-          plate.wells.flat_map do |well|
-            well.used_aliquots.select { |aliquot| aliquot.source_type == 'Pacbio::Library' }
-          end
-        end
-        Emq::Publisher.publish(all_library_aliquots, Pipelines.pacbio, 'volume_tracking')
+        Emq::Publisher.publish(@model.all_library_aliquots, Pipelines.pacbio, 'volume_tracking')
       end
 
       private
--- a/app/resources/v1/pacbio/run_resource.rb
+++ b/app/resources/v1/pacbio/run_resource.rb
@@ -56,6 +56,12 @@ module V1
 
       def publish_messages
         Messages.publish(@model, Pipelines.pacbio.message)
+        all_library_aliquots = @model.plates.flat_map do |plate|
+          plate.wells.flat_map do |well|
+            well.used_aliquots.select { |aliquot| aliquot.source_type == 'Pacbio::Library' }
+          end
+        end
+        Emq::Publisher.publish(all_library_aliquots, Pipelines.pacbio, 'volume_tracking')
       end
 
       private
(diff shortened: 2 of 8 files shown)


app/resources/v1/pacbio/run_resource.rb
```diff
@@ -56,6 +56,12 @@ def self.resource_klass_for(type)
 
       def publish_messages
         Messages.publish(@model, Pipelines.pacbio.message)
+        all_library_aliquots = @model.plates.flat_map do |plate|
```
BenTopping, line 59: all_library_aliquots seems like a method that should be on the model, that way it is easier to test as well.

spec/exchanges/volume_tracking/message_builder_spec.rb
```diff
@@ -51,7 +51,7 @@
                                                           source_type: 'library',
                                                           source_barcode: pacbio_library.tube.barcode,
                                                           sample_name: pacbio_library.sample_name,
-                                                          used_by_type: 'well',
+                                                          used_by_type: 'run',
```
BenTopping, line 54: I haven't been involved in a lot of this work so this might have already been decided but is this right being 'run'? The used_by_barcode is a well barcode so does that need changing?

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -2,27 +2,51 @@
...
-  shared_examples 'publish_messages_on_create' do
+  shared_examples 'publish_messages_on_create' do |aliquot_size|
     it 'publishes a message' do
       expect(Messages).to receive(:publish).with(instance_of(Pacbio::Run), having_attributes(pipeline: 'pacbio'))
       post v1_pacbio_runs_path, params: body, headers: json_api_headers
       expect(response).to have_http_status(:success), response.body
     end
+
+    it 'publishes volume tracking message for each used aliquot' do
+      expect(Emq::Publisher).to receive(:publish).with(array_of_aliquots_with_length(aliquot_size), instance_of(Pipelines::Configuration::Item), 'volume_tracking')
```
BenTopping, line 32: I think we should just be testing that the message is sent correctly here. The contents of the message should be tested by writing tests for the all_library_aliquots function.

BenTopping (changes requested):
Think this is mostly there. Apologies if this has already been discussed elsewhere but is there a reason we are just sending library aliquots and not pool aliquots?

seenanair:
> Think this is mostly there. Apologies if this has already been discussed elsewhere but is there a reason we are just sending library aliquots and not pool aliquots?

There is another [story](https://github.com/sanger/traction-service/issues/1370) coming up for pool ones, that's why I haven't included that inside this. Hope that make sense.

spec/exchanges/volume_tracking/message_builder_spec.rb
seenanair: Good question. In Avro [schema](https://redpanda.uat.psd.sanger.ac.uk/console/schema-registry/subjects/create-aliquot-in-mlwh?version=latest) it has been defined as a run as a source_type, but then the barcode field points to well. I will clarify it with @stevieing

spec/requests/v1/pacbio/runs_spec.rb
BenTopping: Yeah similar to the test above it just checks that a message is received with some data.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -15,6 +15,17 @@
...
+
+    it 'publishes volume tracking message for each used aliquot' do
+      allow(Emq::Publisher).to receive(:publish)
+
+      post v1_pacbio_runs_path, params: body, headers: json_api_headers
+
+      expect(response).to have_http_status(:success), response.body
+      run = Pacbio::Run.first
+      # Verify that the publish method was called with the expected arguments
+      expect(Emq::Publisher).to have_received(:publish).with(array_including(run.all_library_aliquots), instance_of(Pipelines::Configuration::Item), 'volume_tracking') # rubocop:disable RSpec/MessageSpies
```
stevieing, line 27: I think we only need to test that a publish message has been received. I presume the publish message is also tested elsewhere.

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -23,6 +34,16 @@
...
     end
+
+    it 'publishes volume tracking message for each used aliquot' do
+      allow(Emq::Publisher).to receive(:publish)
+
+      patch v1_pacbio_run_path(run), params: body, headers: json_api_headers
+      expect(response).to have_http_status(:success), response.parsed_body
+
+      run = Pacbio::Run.first
+      expect(Emq::Publisher).to have_received(:publish).with(array_including(run.all_library_aliquots), instance_of(Pipelines::Configuration::Item), 'volume_tracking') # rubocop:disable RSpec/MessageSpies
```
stevieing, line 45: As above. 

I think we only need to test that a publish message has been received. I presume the publish message is also tested elsewhere.

app/exchanges/volume_tracking/message_builder.rb
```diff
@@ -33,7 +33,7 @@ def publish_data # rubocop:disable Metrics/MethodLength
 
       case aliquot.used_by_type
       when 'Pacbio::Well'
-        data[:used_by_type] = 'well'
+        data[:used_by_type] = 'run'
         data[:used_by_barcode] =
           "#{aliquot.used_by.plate.sequencing_kit_box_barcode}:#{aliquot.used_by.plate.plate_number}:#{aliquot.used_by.position}" # rubocop:disable Layout/LineLength
```
stevieing, line 38: There seems to be a few of these. You can put the line over 2 lines using "..." \ which means you won't have to disable.

stevieing (changes requested):
A couple of minor things.

spec/requests/v1/pacbio/runs_spec.rb
stevieing: No. That's fine. As long as the test is generic as possible and the publish message is tested correctly elsewhere.

spec/exchanges/volume_tracking/message_builder_spec.rb
seenanair: @BenTopping Just to let you know, I've discussed with Steve and confirmed that it should be 'run' as it makes more sense to external users.

spec/requests/v1/pacbio/runs_spec.rb
seenanair: Yes, it checks for aliquots being passed in as part of the publishing_job_spec, but in a generic way — not specifically for runs created with libraries. I don't think this check to see only library aliquots exists  is necessary, as it will be made redundant by pool aliquots in upcoming stories. Additionally, the content of the message is already covered by message_builder_spec

stevieing (approved):

BenTopping (approved):

Issue #1371, opened 2024-07-26
Y24-217 - As developers we need to push all of the aliquots to the warehouse so that the historical data is available for volume tracking


stevieing:
**User story**
As developers we need to push all of the aliquots to the warehouse so that the historical data is available for volume tracking

**Who are the primary contacts for this story**
Steve / Ben / James

**Who is the nominated tester for UAT**
Steve / Ben / James

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] There should be a reproducible rake task which will push all of the current data to the warehouse
- [ ] All of the volume tracking data should be available in the warehouse

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

PR #1388, opened 2024-08-07
Y24-217 push historic data to warehouse

seenanair:
Closes #https://github.com/sanger/traction-service/issues/1371

- Creates a rake task to push all aliquots to the warehouse using EMQ 

### DIFF
--- a/lib/tasks/push_aliquot_data_to_warehouse.rake
+++ b/lib/tasks/push_aliquot_data_to_warehouse.rake
@@ -7,7 +7,7 @@ namespace :pool_and_library_aliquots do
   task push_data_to_warehouse: :environment do
     puts '-> Pushing all pool and library aliquots data to the warehouse for volume tracking'
 
-    ActiveRecord::Base.transaction do
+    begin
       # Following are the contexts where aliquots are created, out of which only the aliquots from libraries and pools (1,2,3) are used for volume tracking
       # 1. When a library or pool is created, primary aliquot is created
       # 2. When a library is used in pool or run, derived aliquot is created
@@ -18,6 +18,8 @@ namespace :pool_and_library_aliquots do
       # Filter all aliquots that are not from a Pacbio::Request
       aliquots = Aliquot.where.not(source_type: 'Pacbio::Request')
       Emq::Publisher.publish(aliquots, Pipelines.pacbio, 'volume_tracking')
+
+      puts '-> Successfully pushed all pool and library aliquots data to the warehouse'
     end
   end
 end
--- a/lib/tasks/push_aliquot_data_to_warehouse.rake
+++ b/lib/tasks/push_aliquot_data_to_warehouse.rake
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+namespace :pool_and_library_aliquots do
+  #  A rake task to be run to push all pool and library aliquots data to the warehouse for volume tracking
+  desc 'Push all pool and library aliquots data to the warehouse for volume tracking'
+
+  task push_data_to_warehouse: :environment do
+    puts '-> Pushing all pool and library aliquots data to the warehouse for volume tracking'
+
+    ActiveRecord::Base.transaction do
+      # Following are the contexts where aliquots are created, out of which only the aliquots from libraries and pools (1,2,3) are used for volume tracking
+      # 1. When a library or pool is created, primary aliquot is created
+      # 2. When a library is used in pool or run, derived aliquot is created
+      # 3. When a pool is used in run, derived aliquot is created
+      # 4. When a request is imported into the reception, a primary aliquot is created
+      # 5. When a request is used in a pool or library, derived aliquot is created
+
+      # Filter all aliquots that are not from a Pacbio::Request
+      aliquots = Aliquot.where.not(source_type: 'Pacbio::Request')
+      Emq::Publisher.publish(aliquots, Pipelines.pacbio, 'volume_tracking')
+    end
+  end
+end
--- a/lib/tasks/push_aliquot_data_to_warehouse.rake
+++ b/lib/tasks/push_aliquot_data_to_warehouse.rake
@@ -20,6 +20,9 @@ namespace :pool_and_library_aliquots do
       Emq::Publisher.publish(aliquots, Pipelines.pacbio, 'volume_tracking')
 
       puts '-> Successfully pushed all pool and library aliquots data to the warehouse'
+    rescue StandardError => e
+      Rails.logger.error("Failed to publish message: #{e.message}")
+      puts "-> Failed to push aliquots data to the warehouse: #{e.message}"
     end
   end
 end
--- a/lib/tasks/push_aliquot_data_to_warehouse.rake
+++ b/lib/tasks/push_aliquot_data_to_warehouse.rake
@@ -8,15 +8,30 @@ namespace :pool_and_library_aliquots do
     puts '-> Pushing all pool and library aliquots data to the warehouse for volume tracking'
 
     begin
-      # Following are the contexts where aliquots are created, out of which only the aliquots from libraries and pools (1,2,3) are used for volume tracking
+      # Following are the contexts where aliquots are created. We need to push, the primary aliquots created for a pool and library (1,2)
+      # and the derived aliquots of a pool and library (directly or indirectly through a pool) used in a run  to the warehouse
       # 1. When a library or pool is created, primary aliquot is created
       # 2. When a library is used in pool or run, derived aliquot is created
       # 3. When a pool is used in run, derived aliquot is created
       # 4. When a request is imported into the reception, a primary aliquot is created
       # 5. When a request is used in a pool or library, derived aliquot is created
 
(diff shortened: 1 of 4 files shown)


spec/lib/tasks/push_aliquot_data_to_warehouse.rake_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        <<~HEREDOC
+          -> Pushing all pool and library aliquots data to the warehouse for volume tracking
+          -> Successfully pushed all pool and library aliquots data to the warehouse
+        HEREDOC
+      ).to_stdout
+
+      expect(Emq::Publisher).to have_received(:publish).with(array_including(*library_aliquots, *pool_aliquots, *aliquots_used_in_run), instance_of(Pipelines::Configuration::Item), 'volume_tracking') # rubocop:disable RSpec/MessageSpies
+    end
+
+    it 'does not publish Pacbio::Request aliquots to the warehouse' do
```
BenTopping, line 44: Is this test needed? The above test checks that publish is received without request aliquots (even though they exist) so that should be sufficient to show requests aren't included.

lib/tasks/push_aliquot_data_to_warehouse.rake
```diff
@@ -0,0 +1,25 @@
...
+      # Following are the contexts where aliquots are created, out of which only the aliquots from libraries and pools (1,2,3) are used for volume tracking
+      # 1. When a library or pool is created, primary aliquot is created
+      # 2. When a library is used in pool or run, derived aliquot is created
+      # 3. When a pool is used in run, derived aliquot is created
+      # 4. When a request is imported into the reception, a primary aliquot is created
+      # 5. When a request is used in a pool or library, derived aliquot is created
+
+      # Filter all aliquots that are not from a Pacbio::Request
+      aliquots = Aliquot.where.not(source_type: 'Pacbio::Request')
+      Emq::Publisher.publish(aliquots, Pipelines.pacbio, 'volume_tracking')
```
BenTopping, line 20: Could this publish method fail or return error messages?

spec/lib/tasks/push_aliquot_data_to_warehouse.rake_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+require 'rails_helper'
+require 'rake'
+
+# only load Rake tasks if they haven't been loaded already
+Rails.application.load_tasks if Rake::Task.tasks.empty?
+
+RSpec.describe 'RakeTasks' do
+  # Create a default SMRTLink version to be used by factories
+  before do
```
BenTopping, line 11: We have to initialise this smrtlink default all over the place, should definitely move it to a test helper that sets it up on load, probably not in scope for this story.

BenTopping (commented):
Looks sensible, minor comment about testing.

spec/lib/tasks/push_aliquot_data_to_warehouse.rake_spec.rb
seenanair: Yes, I just thought to make it specific, but otherwise, you are right. It is covered as part of another test, so it can be removed.

lib/tasks/push_aliquot_data_to_warehouse.rake
seenanair: Yes, it fails in inner methods and raises an exception under different circumstances, such as when an Avro validation fails or an EMQ connection error occurs. In those cases, the exception propagates up to the publish method. Now, a rescue is added in the call within the rake task to catch the exception. Hope that helps.
[fb721ee](https://github.com/sanger/traction-service/pull/1388/commits/fb712ee66c790e9f41592e67dfaed783d598dd23)
[383218a](https://github.com/sanger/traction-service/pull/1388/commits/383218a1b54cc34c14ff714ff163d6cb080f5f0c)

spec/lib/tasks/push_aliquot_data_to_warehouse.rake_spec.rb
seenanair: Yes, it is inconvenient to do it in every single place a run is created. Probably make sense to add as a separate story, I hope.

lib/tasks/push_aliquot_data_to_warehouse.rake
dasunpubudumal: I assume `Aliquot.where.not` call returns a list of aliquots. Have we tested publishing a list of aliquots to the EMQ? I can't remember whether `tol-lab-share` is able to decode lists of messages.

seenanair: It returns an array, but these are not the direct arguments sent to EMQ. This is a wrapper call which calls other objects that convert these aliquots into the required message format, which will then be published one by one. Hope that makes sense.

spec/lib/tasks/push_aliquot_data_to_warehouse.rake_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    it 'logs and prints an error message when publishing fails' do
+      error_message = 'Test error'
+      allow(Emq::Publisher).to receive(:publish).and_raise(StandardError.new(error_message))
+
+      expect { Rake::Task['pool_and_library_aliquots:push_data_to_warehouse'].invoke }.to output(
+        a_string_including("-> Failed to push aliquots data to the warehouse: #{error_message}")
+      ).to_stdout
+
+      expect(Rails.logger).to have_received(:error).with("Failed to publish message: #{error_message}") # rubocop:disable RSpec/MessageSpies
```
BenTopping, line 54: If you move this method above the rake task invocation (to line 49) and change it to `expect(Rails.logger).to receive(:error).with("Failed to publish message: #{error_message}")` you can remove this message spies rubocop comment. This change moves it to be a predictive expectation.

spec/lib/tasks/push_aliquot_data_to_warehouse.rake_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      expect(Emq::Publisher).to have_received(:publish).with(array_including(*library_aliquots, *pool_aliquots, *aliquots_used_in_run), instance_of(Pipelines::Configuration::Item), 'volume_tracking') # rubocop:disable RSpec/MessageSpies
+    end
+
+    it 'logs and prints an error message when publishing fails' do
+      error_message = 'Test error'
+      allow(Emq::Publisher).to receive(:publish).and_raise(StandardError.new(error_message))
+
+      expect { Rake::Task['pool_and_library_aliquots:push_data_to_warehouse'].invoke }.to output(
+        a_string_including("-> Failed to push aliquots data to the warehouse: #{error_message}")
```
BenTopping, line 51: Can we change this to heredoc same as above.

BenTopping (changes requested):

lib/tasks/push_aliquot_data_to_warehouse.rake
dasunpubudumal: Yeah that makes sense 😄 Thanks.

dasunpubudumal: This is actually a very good opportunity to test the robustness of our messaging platform and our overall volume tracking solution. I suppose we could use training data, and check the UAT aliquots table since the training instance of our message consumer is pointed to UAT (mlwh doesn't have a training instance).

dasunpubudumal (approved):
This is actually a very good opportunity to test the robustness of our messaging platform and our overall volume tracking solution. I suppose we could use training data, and cross-check the UAT mlwh aliquots table since the training instance of our message consumer is pointed to UAT mlwh (mlwh doesn't have a training instance).

seenanair:
> This is actually a very good opportunity to test the robustness of our messaging platform and our overall volume tracking solution. I suppose we could use training data, and cross-check the UAT mlwh aliquots table since the training instance of our message consumer is pointed to UAT mlwh (mlwh doesn't have a training instance).
Yes, I have dumped the training data to the local database and tested it locally, which seems to be working fine, at least checked until the point where it is successfully publishing to mlwh consumer queue. 
While testing it in uat, we can check whether the other end works as well so it updates the mlwh table.

sabrine33 (approved):

BenTopping (approved):

dasunpubudumal:
> > This is actually a very good opportunity to test the robustness of our messaging platform and our overall volume tracking solution. I suppose we could use training data, and cross-check the UAT mlwh aliquots table since the training instance of our message consumer is pointed to UAT mlwh (mlwh doesn't have a training instance).
> 
> Yes, I have dumped the training data to the local database and tested it locally, which seems to be working fine, at least checked until the point where it is successfully publishing to mlwh consumer queue. While testing it in uat, we can check whether the other end works as well so it updates the mlwh table properly.

Nice. We can check full end to end process if we do this in training. We will have to do some sort of cross validation to check the consistency of the data between the aliquots in traction and the ones persisted in mlwh.

stevieing (approved):

BenTopping (approved):

PR #1389, opened 2024-08-07
sending the pool volume to the warehouse for tracking purposes

sabrine33:


app/exchanges/volume_tracking/message_builder.rb
```diff
@@ -13,35 +13,57 @@ class MessageBuilder < Message::Message
...
+      case object.source_type
       when 'Pacbio::Library'
         data[:source_type] = 'library'
-        data[:source_barcode] = aliquot.source.tube.barcode
-        data[:sample_name] = aliquot.source.sample_name
+        data[:source_barcode] = object.source.tube.barcode
+        data[:sample_name] = object.source.sample_name
+      when 'Pacbio::Pool'
+        data[:source_type] = 'pool'
+        data[:source_barcode] = object.source.tube.barcode
```
seenanair, line 49: Do we need to fill in sample name here? Good to ask @stevieing - with multiple samples in case of a pool, may be string which joins a friendly name for all?

app/resources/v1/pacbio/pool_resource.rb
```diff
@@ -73,7 +76,11 @@ def updated_at
         @model.updated_at.to_fs(:us)
       end
 
-      def publish_messages
+      def publish_messages_on_creation
+        Emq::Publisher.publish(@model.primary_aliquot, Pipelines.pacbio, 'volume_tracking')
+      end
+
+      def publish_messages_on_update
         Messages.publish(@model.sequencing_runs, Pipelines.pacbio.message)
```
seenanair, line 84: I think we need to do the volume_tracking publishing here as well after an update and the same comment as above for library aliquots used in pool.

app/resources/v1/pacbio/pool_resource.rb
```diff
@@ -73,7 +76,11 @@ def updated_at
         @model.updated_at.to_fs(:us)
       end
 
-      def publish_messages
+      def publish_messages_on_creation
+        Emq::Publisher.publish(@model.primary_aliquot, Pipelines.pacbio, 'volume_tracking')
```
seenanair, line 80: Along with the primary_aliquot, I think, if  libraries are used while creating a pool, those aliquots from all libraries that are used in the pool also need to be sent for volume tracking.

spec/models/pacbio/run_spec.rb
```diff
@@ -300,14 +300,12 @@
...
-    it 'only returns aliquots with source_type Pacbio::Library' do
+    it 'returns all the aliquots used for a run' do
       library = create(:pacbio_library)
       pools = create_list(:pacbio_pool, 2)
       wells = [build(:pacbio_well, row: 'A', column: '1', libraries: [library]), build(:pacbio_well, row: 'B', column: '1', pools:)]
       run = create(:pacbio_revio_run, plates: [build(:pacbio_plate, wells:)])
-      run.all_library_aliquots.each do |aliquot|
-        expect(aliquot.source_type).to eq('Pacbio::Library')
-      end
+      expect(run.used_aliquots.size).to eq 8
```
seenanair, line 308: This test for `all_library_aliquots` can remain, I suppose. Otherwise, the `all_library_aliquots` method in the run model would be left untested. Alternatively, if it is not used anywhere, all_library_aliquots can be removed.

seenanair (commented):
Looking good, just shared a few thoughts.

app/exchanges/volume_tracking/message_builder.rb
sabrine33: I was expected this comment, as I am sure, the object does not actually contain a sample name property @stevieing any inputs in this, please

app/resources/v1/pacbio/pool_resource.rb
sabrine33: I am not sure, but also I am not sure what you by those aliquots (from all libraries that are used in the pool) are you referring  to the used_aliquots_attributes objects ?

spec/models/pacbio/run_spec.rb
sabrine33: I actually updated the all_library_aliquots method to used_aliquots as we need to publish for aliquots with source type = Pacbio::Pool too. 
This explains the updates test

app/resources/v1/pacbio/pool_resource.rb
seenanair: I mean  the `used_aliquots` within a pool which can be from a library or a request. As I mentioned, my idea is that the used_aliquots in a pool which is having source_type as library should also be published for volume_tracking, but good to confirm.

app/exchanges/volume_tracking/message_builder.rb
stevieing: A pool will have multiple samples through libraries so yes we could get the names of the samples in the libraries.

app/resources/v1/pacbio/pool_resource.rb
stevieing: We should send the primary aliquot when we create a pool as with libraries. We should then send the used aliquots when the run is created. From what I understand the libraries used aliquots are already sent. It might be worth having a quick meeting to go through this along with @BenTopping just to make sure we have everything covered.

app/exchanges/volume_tracking/message_builder.rb
sabrine33: is so how are we storing that in the table, are concatenating the sample names in one string (join by , maybe )?

stevieing: > is so how are we storing that in the table, are concatenating the sample names in one string (join by , maybe )?

Yes. We can join the with a comma I guess.

seenanair (approved):
Looking good 👍

sabrine33:
> Looking good 👍

Thank you, I would not make it without your help :)

PR #1397, opened 2024-08-14
send the used aliquots on a run

sabrine33:


app/models/pacbio/run.rb
```diff
@@ -99,10 +99,19 @@ def update_smrt_link_options(options)
 
     # Returns all the used aliquots
     # @return [Array<Aliquot>] an array of the used aliquots in a run
-    def used_aliquots
+    def aliquots_to_publish_on_run
+      to_publish = []
       plates.flat_map do |plate|
-        plate.wells.flat_map(&:used_aliquots)
+        source_aliquots = plate.wells.flat_map(&:used_aliquots).select do |aliquot|
```
seenanair, line 105: Is it possible to give more intuitive name for source_aliquots - used_aliquots_with_pool_source?

app/models/pacbio/run.rb
```diff
@@ -99,10 +99,19 @@ def update_smrt_link_options(options)
...
+          aliquot.source_type == 'Pacbio::Pool'
+        end
+        lib_used_aliquots = source_aliquots.flat_map(&:source).flat_map(&:used_aliquots)
+                                           .select do |aliquot|
+          aliquot.source_type == 'Pacbio::Library'
+        end
+        to_publish.concat(source_aliquots, lib_used_aliquots)
```
seenanair, line 112: Additionally, it would be helpful to add some documentation here to clarify what we are trying to accomplish, as it could be confusing for someone picking it up later.

spec/models/pacbio/run_spec.rb
```diff
@@ -300,12 +300,15 @@
   end
 
   describe 'all_library_aliquots' do
```
seenanair, line 302: Does this need to be changed to aliquots_to_publish_on_run

spec/models/pacbio/run_spec.rb
```diff
@@ -300,12 +300,15 @@
...
-      wells = [build(:pacbio_well, row: 'A', column: '1', libraries: [library]), build(:pacbio_well, row: 'B', column: '1', pools:)]
+    it 'returns all the aliquots used for a run of source type eq to library or pool' do
+      request_pool = create(:pacbio_pool, primary_aliquot: create(:aliquot, source: build(:pacbio_request), aliquot_type: :primary))
+      pool = create(:pacbio_pool)
+      wells = [build(:pacbio_well, row: 'A', column: '1', libraries: [create(:pacbio_library)]),
+               build(:pacbio_well, row: 'B', column: '1', pools: [request_pool, pool])]
       run = create(:pacbio_revio_run, plates: [build(:pacbio_plate, wells:)])
-      expect(run.used_aliquots.size).to eq 8
+      run.aliquots_to_publish_on_run do |aliquot|
+        expect(aliquot.source_type).to eq('Pacbio::Library').or eq('Pacbio::Pool')
```
seenanair, line 310: Also, it would be good to add a test with a pool with library in, to check the specific case that we have added now and to see run.aliquots_to_publish_on_run includes the aliquot for that library in pool.

seenanair (approved):

Issue #1403, opened 2024-08-16
Y24-262 - As developers we would like to persist the label printing label workflows so we can update them without having to redeploy traction


stevieing:
**User story**
As developers we would like to persist the label printing label workflows so we can update them without having to redeploy traction

**Who are the primary contacts for this story**
Steve

**Who is the nominated tester for UAT**
PSD

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] create a workflow table which should include the pipeline
- [ ] create a workflow step table which should include the suffix and description

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

PR #1428, opened 2024-09-11
Y24-262 persist label printing workflows

seenanair:
Closes https://github.com/sanger/traction-service/issues/1403

-  Migrations to create workflow and workflow_step tables
- Added Workflow and WorkflowStep models
- Added Workflow and WorkflowStep resources
- Added routes for Workflow
- Added a rake task for creating and updating workflows
- Added a WorkflowPipelineable concern, as the pipelines added in workflows differ from the existing Pipelinable concern, which is used in many other places, keeping the two concerns separate

     


app/models/concerns/workflow_pipelineable.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Include in records to associate them with a pipeline via an enum
+module WorkflowPipelineable
+  extend ActiveSupport::Concern
```
stevieing, line 5: I think it is better to use pipelineable? Not sure of the changes we need to make.

stevieing (changes requested):
Only one issue which is the use of pipelineable.

app/models/concerns/workflow_pipelineable.rb
seenanair: Thanks for the review. 
I was considering the WorkflowPipelineable concern, as the pipelines added in workflows differ from the existing Pipelinable concern, which is used in many other places. 
Thought it would be better to keep the two concerns separate as the workflow enum pipeline: { pacbio: 0, ont: 1, extraction: 2, sample_qc: 3, hic: 4, bio_nano: 5 }. has only two entries in common with Pipelinable—pacbio and ont. 
Also, I was thinking whether changing this enum would cause issues with Pipelinable, since it is used in many other places.

stevieing (approved):

PR #1432, opened 2024-09-17
misc: 

BenTopping:
Rails 7.2 changed default active record attributes inspection to just display id's with returned objects.
This is unhelpful for Neil when debugging issues and working on data fixes.

- Updates config to always show all attributes when inspecting elements.


BenTopping:
Closing as this behaviour is being reverted in the future rails release so this PR is no longer required.

PR #1442, opened 2024-09-26
Y24-320 - modify sample sheet behaviour so that Revio sample sheets behave as untagged when the behaviour is set to hidden

stevieing:
Closes #

- 

### DIFF
--- a/app/models/aliquot.rb
+++ b/app/models/aliquot.rb
@@ -50,7 +50,10 @@ class Aliquot < ApplicationRecord
   #
   # @return [Boolean] Returns true if the aliquot is tagged, false otherwise.
   def tagged?
-    tag.present?
+    # This feels like a bit of a hack but I wasn't exactly sure where the best place to
+    # it. I tried to follow the sample sheet behaviour but got lost.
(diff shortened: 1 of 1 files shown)


BenTopping (approved):
I think this is probably fine but it does mean `tagged?` is no longer just whether the aliquot has a tag or not which might be confusing.

I haven't really looked into this so ignore if irrelevant but could you do something in the [sample sheet aliquot concern](https://github.com/sanger/traction-service/blob/develop/app/models/concerns/sample_sheet.rb#L120)

stevieing:
> I think this is probably fine but it does mean `tagged?` is no longer just whether the aliquot has a tag or not which might be confusing.
> 
> I haven't really looked into this so ignore if irrelevant but could you do something in the [sample sheet aliquot concern](https://github.com/sanger/traction-service/blob/develop/app/models/concerns/sample_sheet.rb#L120)

I had a look. There is a lot of indirection and obfuscation so there were a few different methods which led to checking whether there is a tag and I couldn't quite work it out. I have created a story #1443 to refactor it.

PR #1454, opened 2024-10-11
Proposing changing `created_at` format to include seconds, to fix library order

harrietc52:
Closes https://github.com/sanger/traction-ui/issues/1993

- The idea is to change the format of the `created_at` data, returned to the Traction UI from the Traction Service, so that it includes the seconds `yyyy-mm-dd hh:mm:ss`. This is to help solve the issue where two libraries are created within the same minute but the ordering is not predictably by the latest, resulting in barcodes being swapped. 
- If this does solve the problem, maybe updating timestamps through out the application to include seconds could prevent similar issues for other resources. (Changing from `.to_fs(:us)` to `.to_fs(:db)`). I also can’t seem to find any documentation on  `.to_fs(:us)`, so not sure if that format is something we’re specifying?
- The new format would be e.g. `2024-10-11 13:04:47`, which is different to the old format `2024/10/11 12:50`. We would need to check with the users if this is okay, or possibly reformat it in the front end to match the old, or, if we do decide to update them all to include seconds, maybe them all being consistent with the new format is okay

PacBio Libraries Page:

Before 
Screenshot 2024-10-11 at 14 49 38

After 
Screenshot 2024-10-11 at 14 50 49

- Start up Traction Service and UI
- From PacBio Sample page, create two libraries within the same minute
- Go to PacBio Library page, and they should be ordered by default with the most recent at the top
- The UI should show the seconds.

### DIFF
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -121,7 +121,8 @@ module V1
       end
 
       def created_at
-        @model.created_at.to_fs(:us)
+        # to_fs(:db) -> yyyy-mm-dd hh:mm:ss
+        @model.created_at.to_fs(:db)
       end
 
       def deactivated_at
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -121,8 +121,7 @@ module V1
       end
 
       def created_at
-        # to_fs(:db) -> yyyy-mm-dd hh:mm:ss
-        @model.created_at.to_fs(:db)
+        @model.created_at.to_fs(:us)
       end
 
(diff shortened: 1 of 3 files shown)


BenTopping (approved):
Looks good, nice solution.

BenTopping:
Just thinking about this a bit further based on Steve's message in the traction slack - If we are bulk creating libraries they could be created in the same second which may cause issues as well. Not sure if thats worth considering here or not.

stevieing:
> Just thinking about this a bit further based on Steve's message in the traction slack - If we are planning on bulk creating libraries they could be created in the same second which may cause issues as well. Not sure if thats worth considering here or not.

It is worth considering. This is partly why I wanted to use the id in reverse order.

PR #1455, opened 2024-10-14
Publishing filtered aliquots

### LINKED ISSUE #1451: Y24-370 - As Tree of Life we would like the sample data in the warehouse when a library is created so that we can track back to the original plate or tube barcode to make the top up process more efficient
**User story**
As Tree of Life we would like the sample data in the warehouse when a library is created so that we can track back to the original plate or tube barcode to make the top up process more efficient.

Ruth / Caroline / Steve

PSD / Caroline

**Acceptance criteria**
- [x] When a library is created the sample should be pushed to the aliquot table
- [x] The sample used_by should be the library
- [x] The sample should have a source barcode which is the plate or tube barcode

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

dasunpubudumal:
Closes #1451 

- Filter request aliquots and publish them to mlwh.

### DIFF
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -65,7 +65,6 @@ module VolumeTracking
                                 else
                                   object.source.tube.barcode
                                 end
-        data[:source_barcode] = object.source.container.barcode
         data[:sample_name] = object.source.sample_name
       end
     end
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -60,11 +60,7 @@ module VolumeTracking
         data[:sample_name] = pacbio_library_sample_names
       when 'Pacbio::Request'
         data[:source_type] = 'request'
-        data[:source_barcode] = if object.source.tube.nil?
-                                  object.source.plate.barcode
-                                else
-                                  object.source.tube.barcode
-                                end
+        data[:source_barcode] = object.source.container.barcode
         data[:sample_name] = object.source.sample_name
       end
     end
@@ -81,6 +77,7 @@ module VolumeTracking
         data[:used_by_barcode] = object.used_by.tube.barcode
       when 'Pacbio::Library'
         data[:used_by_type] = 'library'
+        data[:used_by_barcode] = object.used_by.tube.barcode
       end
     end
 
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -79,8 +79,8 @@ module VolumeTracking
       when 'Pacbio::Pool'
         data[:used_by_type] = 'pool'
         data[:used_by_barcode] = object.used_by.tube.barcode
-      when 'Pacbio::Request'
-        data[:used_by_type] = 'request'
+      when 'Pacbio::Library'
+        data[:used_by_type] = 'library'
       end
     end
 
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -56,6 +56,15 @@ module VolumeTracking
         data[:source_type] = 'pool'
         data[:source_barcode] = object.source.tube.barcode
         data[:sample_name] = pacbio_library_sample_names
+      when 'Pacbio::Request'
+        data[:source_type] = 'request'
+        if object.source.tube.nil?
+          data[:source_barcode] = object.source.plate.barcode
+        else
+          data[:source_barcode] = object.source.tube.barcode
+        end
+        data[:source_barcode] = object.source.container.barcode
+        data[:sample_name] = object.source.sample_name
       end
     end
 
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -46,6 +46,8 @@ module VolumeTracking
       }
     end
 
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
     def populate_by_source_type(data)
       case object.source_type
       when 'Pacbio::Library'
@@ -67,6 +69,8 @@ module VolumeTracking
         data[:sample_name] = object.source.sample_name
       end
     end
+    # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/MethodLength
 
     def populate_by_used_type(data)
       case object.used_by_type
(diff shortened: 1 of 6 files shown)


BenTopping (changes requested):

app/exchanges/volume_tracking/message_builder.rb
```diff
@@ -56,8 +58,19 @@ def populate_by_source_type(data)
         data[:source_type] = 'pool'
         data[:source_barcode] = object.source.tube.barcode
         data[:sample_name] = pacbio_library_sample_names
+      when 'Pacbio::Request'
+        data[:source_type] = 'request'
+        data[:source_barcode] = if object.source.tube.nil?
+                                  object.source.plate.barcode
```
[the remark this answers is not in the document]
dasunpubudumal: The requirements say that "he sample should have a source barcode which is the plate or tube barcode" so I assumed the plate or tube barcode would be enough?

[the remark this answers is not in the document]
BenTopping: Ok, yeah just wanted to double check 👌

app/exchanges/volume_tracking/message_builder.rb
```diff
@@ -56,8 +58,19 @@ def populate_by_source_type(data)
         data[:source_type] = 'pool'
         data[:source_barcode] = object.source.tube.barcode
         data[:sample_name] = pacbio_library_sample_names
+      when 'Pacbio::Request'
+        data[:source_type] = 'request'
+        data[:source_barcode] = if object.source.tube.nil?
```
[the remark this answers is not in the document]
BenTopping: If line 68 was working (using container) I think that is nicer than doing an if check

app/resources/v1/pacbio/library_resource.rb
```diff
@@ -130,11 +130,12 @@ def deactivated_at
...
       def publish_messages
         Messages.publish(@model.sequencing_runs, Pipelines.pacbio.message)
-        publish_volume_tracking_message
+        publish_volume_tracking_messages
       end
 
-      def publish_volume_tracking_message
-        Emq::Publisher.publish(@model.primary_aliquot, Pipelines.pacbio, 'volume_tracking')
+      def publish_volume_tracking_messages
+        Emq::Publisher.publish(Array(@model.primary_aliquot) + Array(@model.used_aliquots),
```
BenTopping, line 137:
```suggestion
        Emq::Publisher.publish([@model.primary_aliquot, *@model.used_aliquots]),
```

spec/exchanges/volume_tracking/message_builder_spec.rb
```diff
@@ -91,6 +91,24 @@
...
+      let(:message_builder) { described_class.new(object: aliquot, configuration:) }
+      let(:pacbio_request) { create(:pacbio_request, tube: create(:tube, barcode: '123456789')) }
+
+      it 'produces the message in the correct format' do
+        expect(message_builder.publish_data).to include({
+                                                          source_type: 'request',
+                                                          source_barcode: pacbio_request.tube.barcode,
+                                                          sample_name: pacbio_request.sample_name,
+                                                          used_by_type: 'library',
+                                                          used_by_barcode: '',
```
BenTopping, line 106: Should the barcode here be the library barcode?

app/exchanges/volume_tracking/message_builder.rb
```diff
@@ -56,8 +57,13 @@ def populate_by_source_type(data)
         data[:source_type] = 'pool'
         data[:source_barcode] = object.source.tube.barcode
         data[:sample_name] = pacbio_library_sample_names
+      when 'Pacbio::Request'
+        data[:source_type] = 'request'
+        data[:source_barcode] = object.source.container.barcode
```
seenanair, line 62: Oh,  Is there a source.container accessor? Good way to get source barcode which avoids checking whether it's a tube or a plate. 👍

config/pipelines/pacbio.yml
```diff
@@ -304,7 +304,7 @@ default: &default
     # The number suffix '1' is used to identify the schema version of message (created by the
     # message builder class defined in the message_class key) 
     # The schema version number should be present in bunny.yml for the same 'key'
-      avro_schema_version_1 : 
+      avro_schema_version_1: &avro_schema_version_1
```
seenanair, line 307: Do we need to keep this version as  only one version will picked up from bunny.yml  ?

config/pipelines/pacbio.yml
```diff
@@ -351,6 +351,8 @@ default: &default
           usedByType:
             type: :self
             value: publish_data&.used_by_type
+      avro_schema_version_2:
+        <<: *avro_schema_version_1
```
seenanair, line 355: One doubt, Is the latest avro_schema_version is v '2' or '3'?

dasunpubudumal: I think it's 2. Previously it was 1.

config/pipelines/pacbio.yml
dasunpubudumal: I'm not entirely sure. The tests are highly coupled with this key. For example, if I remove `avro_schema_version_1`, I would have to change all the tests as well. 

My understanding of this `yml` was that we would keep the message building configs corresponding to each schema in this file. Not sure if I'm correct or incorrect in that regard.

seenanair: Ok, it's not a big issue for now, but something to think about in future. 
I think the dependency with the tests is because we are keeping a stub response at spec/fixtures/volume_tracking_avro_response.json which is for version 1.

seenanair (approved):

app/exchanges/volume_tracking/message_builder.rb
dasunpubudumal: Courtesy of @BenTopping 😄

config/pipelines/pacbio.yml
dasunpubudumal: Yeah that's correct. But because we have the schemas in RedPanda, and we are _not_ removing them, if we decide to revert back to an older schema, that mean we would have to change this `pacbio.yml` file if we remove the older keys (e.g., `avro_schema_version_1`). 

I always thought this part of the `pacbio.yml` file as a versioned history of the mapping between the schema and the message builder.

app/exchanges/volume_tracking/message_builder.rb
seenanair: @BenTopping 🔥 😄

BenTopping: Technically the container is either a well or a tube, but well's delegate barcode to their plate. So you can use it to retrieve the barcode.

BenTopping (approved):

PR #1475, opened 2024-10-30
Y24-406 - create SMRT Link 13.1 sample sheets

sabrine33:


config/pacbio_smrt_link_versions.yml
```diff
@@ -36,7 +36,10 @@ default: &default
       name: v13_sequel_iie
       active: true
       default: false
-
+    v13_1_sequel_iie:
```
stevieing, line 39: We will also need a 13.1 version for Revio. Although there are no changes it is best for consistency.

stevieing (changes requested):

stevieing (approved):
I haven't run this locally but hopefully the sample sheets are working correctly?

Once we deploy to UAT we can upload it to the test SMRT Link Instance.

sabrine33:
> I haven't run this locally but hopefully the sample sheets are working correctly?
> 
> Once we deploy to UAT we can upload it to the test SMRT Link Instance.

Yes, I ran a few tests. The CSV version column is at the beginning of the file, with a value of 1 that only gets populated for the first row. This is also confirmed by the unit tests.

PR #1479, opened 2024-11-06
Y24-411 - PacBio batch library creation

### LINKED ISSUE #1461: Y24-411 - As a long read user (Michelle) I would like to be able to create multiple libraries at once (24) to save time and reduce errors
**User story**
As a long read user (Michelle) I would like to be able to create multiple libraries at once (24) to save time and reduce errors

Steve / Michelle

Michelle / Long Read

**Acceptance criteria**
- [ ] The libraries endpoint should be able to receive a request to create multiple libraries OR a new endpoint for creating multiple libraries
- [ ] each library should be validated as a standard library
- [ ] the libraries should be returned with the newly created tubes so the barcodes can be printed

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

BenTopping:
Closes #1461 

- Adds LibraryBatch model - a similar concept to the reception to help track and create groups of libraries.
- Adds LibraryBatch resource and api routes.


app/resources/v1/pacbio/library_batch_resource.rb
```diff
@@ -0,0 +1,68 @@
...
+    #   }
+    #
+    # @note Access this resource via the `/v1/pacbio/library_batches/` endpoint.
+    #
+    # Provides a JSON:API representation of {Pacbio::LibraryBatch}.
+    #
+    # For more information about JSON:API see the [JSON:API Specifications](https://jsonapi.org/format/)
+    # or look at the [JSONAPI::Resources](http://jsonapi-resources.com/) package
+    # for the service implementation of the JSON:API standard.
+    class LibraryBatchResource < JSONAPI::Resource
```
stevieing, line 39: I think we may need to add some relationships here? We will need to return the tube barcodes and also add that into the documentation.

stevieing (changes requested):
Looking good. Good decision to use batches as it reduces complexity in the library endpoint.

I think we need to return the tubes so users can print the barcodes.

app/resources/v1/pacbio/library_batch_resource.rb
BenTopping: There is a test case for this. When doing a post you can include relationships, if we do post with `?include='libraries.tube'` we will get the tubes back.

stevieing: Ok. That's great but could we document it in the resource? 

I am talking to Kieron Taylor at the moment and he would like to start using our endpoints but the docs are sketchy or missing so having something explicit in there would help a lot. I think we will also start adding it to other resources.

e.g. I am looking at pacbio pooling and I have found how to retrieve the includes and filter but did that by looking at the ui, resource and network in developer tools. I think having these things at the top of the resource will make it much easier.

As part of my story I plan to add it for this example.

BenTopping: Ah I see. Ok sure ill add a comment on it.

stevieing (approved):

PR #1486, opened 2024-11-18
Y24-403 - Send run name with well library aliquots

### LINKED ISSUE #1457: Y24-403 - As NPG we would like the run name sent to the warehouse when a library aliquot is used in a run so that we can use it to link it to a run so we can find the insert size of the library to allow the qc team to use it in their analysis
**User story**
As NPG we would like the run name sent to the warehouse when a library aliquot so that we can use it to link it to a run so we can find the insert size of the library to allow the qc team to use it in their analysis

Steve / Marina

Steve / Marina

**Acceptance criteria**
- [ ] when a library is used on a run and the aliquot is sent to the warehouse it should be a concatentation of the run_name and well_location

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**
follow on from #811941: Propagate library size to pac_bio_run table (MLWH)

BenTopping:
Closes #1457

- Updates volume tracking message builder to send used_by_well_barcode in the format <run_name>:<well_position>

Wondering if the plate number is still useful?

### DIFF
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -80,7 +80,7 @@ module VolumeTracking
     end
 
     def used_by_well_barcode
-      "#{object.used_by.run.name}:#{object.used_by.position}"
+      "#{object.used_by.run.name}:#{object.used_by.plate.plate_number}:#{object.used_by.position}"
     end
 
     def pacbio_library_sample_names
--- a/app/exchanges/volume_tracking/message_builder.rb
+++ b/app/exchanges/volume_tracking/message_builder.rb
@@ -80,8 +80,7 @@ module VolumeTracking
     end
 
     def used_by_well_barcode
-      "#{object.used_by.plate.sequencing_kit_box_barcode}:" \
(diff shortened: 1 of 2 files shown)


app/exchanges/volume_tracking/message_builder.rb
```diff
@@ -80,8 +80,7 @@ def populate_by_used_type(data)
     end
 
     def used_by_well_barcode
-      "#{object.used_by.plate.sequencing_kit_box_barcode}:" \
-        "#{object.used_by.plate.plate_number}:#{object.used_by.position}"
+      "#{object.used_by.run.name}:#{object.used_by.position}"
```
stevieing, line 83: Apologies. It looks like I got this slightly wrong in the story. We will still need the plate number otherwise we would not be able to identify them uniquely.

BenTopping: Yeah makes sense, will add it back 👍

stevieing (approved):

seenanair (approved):

PR #1494, opened 2024-11-27
fix(primary-aliquot): fixes bug where pool and library primary aliquots without ids caused the aliquots to be remade

BenTopping:

- The UI can send primary aliquots of existing pools/libraries without their ids. This change prevents the aliquot being deleted and recreated in that case.

### DIFF
--- a/config/bunny.yml.example
+++ b/config/bunny.yml.example
@@ -21,7 +21,7 @@ default: &default
       subjects:
         volume_tracking:
           subject: 'create-aliquot-in-mlwh'
-          version: 3
+          version: 2
 
 development:
   <<: *default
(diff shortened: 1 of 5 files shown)


stevieing (approved):

config/bunny.yml.example
```diff
@@ -21,7 +21,7 @@ default: &default
       subjects:
         volume_tracking:
           subject: 'create-aliquot-in-mlwh'
-          version: 3
+          version: 2
```
seenanair, line 24: Just to clarify, why did the Red Panda avro schema version change?

BenTopping: Because there is no version 3. I spoke to Dasun and he said he probably left it at 3 while he was testing.

seenanair (approved):

Issue #1490, opened 2024-11-22
Y24-473 - Extract and generalize multiple source_identifier filtering logic  


seenanair:
**Describe the Housekeeping**

The source_identifier filter logic currently implemented in the request_resource supports passing multiple values against a single field, enabling efficient matching of multiple barcodes in a single request and retrieving all records that match the valid barcodes.
https://github.com/sanger/traction-service/blob/9df3648ee8c9afcec02c989ea8529491f314c3bb/app/resources/v1/pacbio/request_resource.rb#L63

Extracting this logic from the request_resource and refactoring it into a shared utility or module will allow this functionality to be reused across all resources. This approach ensures consistency, reduces code duplication, and improves maintainability.

The extracted filter should be designed to be flexible, accommodating the specific requirements of each resource while maintaining a standardized  approach to barcode filtering.

**Blocking issues**
Describe any other issues or tickets that may be blocking this change.

**Additional context**
https://github.com/sanger/traction-service/pull/1488#pullrequestreview-2454409961_

PR #1488, opened 2024-11-22
Y24-412 - request end point filter update with multiple source identifiers

seenanair:
Closes https://github.com/sanger/traction-service/issues/1489

#### Changes proposed in this pull request
-Enhanced the `filter :source_identifier` in request_resource to support multiple source identifiers instead of just one.
-Added filter by name option to pacbio tagsets resource

#### Instructions for Reviewers

_[All PRs] - Confirm PR template filled_  
_[Feature Branches] - Review code_  
_[Production Merges to `main`]_  
 &nbsp; &nbsp; \- _Check story numbers included_  
 &nbsp; &nbsp; \- _Check for debug code_  
 &nbsp; &nbsp; \- _Check version_

PR #1508, opened 2024-12-11
Y24 473 generalize multiple source identifier filtering logic

seenanair:
Closes https://github.com/sanger/traction-service/issues/1490

Added the SourceIdentifierFilterable concern, thereby generalizing the code to apply multiple source_identifiers, which can be called from both Pacbio::LibraryResource and Pacbio::RequestResource.
https://github.com/sanger/traction-service/pull/1488#pullrequestreview-2454409961

### DIFF
--- a/app/models/concerns/source_identifier_filterable.rb
+++ b/app/models/concerns/source_identifier_filterable.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# app/models/concerns/barcode_filterable.rb
+module SourceIdentifierFilterable
+  extend ActiveSupport::Concern
+
+  class_methods do # rubocop:disable Metrics/BlockLength
+    def apply_source_identifier_filter(records, value, # rubocop:disable Metrics/MethodLength,Metrics/PerceivedComplexity
+                                       plate_join: :plate, tube_join: :tube, well_join: :well)
+      rec_ids = []
+      value.each do |val|
+        # byebug
+        if val.include?(':')
+          plate, well = val.split(':')
+          if plate.present?
+            filtered_recs = records.joins(plate_join).where(plate_join => { barcode: plate })
+            if well.present?
+              filtered_recs = filtered_recs.joins(well_join).where(well_join => { position: well })
+            end
+          else
+            Rails.logger.warn("Malformed source identifier: '#{val}'. Plate part is missing.")
+            next
+          end
+        else
+          filtered_recs = records.joins(plate_join).where(plate_join => { barcode: val })
+          if filtered_recs.empty?
+            filtered_recs = records.joins(tube_join).where(tube_join => { barcode: val })
+          end
+        end
+        rec_ids.concat(filtered_recs.pluck(:id))
+      rescue StandardError => e
+        Rails.logger.warn("Invalid source identifier: #{val}, error: #{e.message}")
+      end
+      records.where(id: rec_ids)
+    end
+  end
+end
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -15,6 +15,7 @@ module V1
     # for the service implementation of the JSON:API standard.
     class LibraryResource < JSONAPI::Resource
       include Shared::RunSuitability
+      include SourceIdentifierFilterable
 
       model_name 'Pacbio::Library'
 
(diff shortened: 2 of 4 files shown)


stevieing (commented):
Looks really good. With plenty of documentation.

I will have a proper look later. Are there any specific filters I can use if I run locally to test it?

seenanair:
> Looks really good. With plenty of documentation.
> 
> I will have a proper look later. Are there any specific filters I can use if I run locally to test it?
Thanks.
 Yes, it is for the 'source' filter. To test it, we need to select 'Source' in the Libraries and Samples page within the 'Filter Results' panel, using multiple values of source separated by commas.

app/models/concerns/source_identifier_filterable.rb
```diff
@@ -0,0 +1,73 @@
...
+#                                                             plate_join: :source_plate,
+#                                                             tube_join: :source_tube)
+#
+module SourceIdentifierFilterable
+  extend ActiveSupport::Concern
+
+  class_methods do # rubocop:disable Metrics/BlockLength
+    def apply_source_identifier_filter(records, value, # rubocop:disable Metrics/MethodLength,Metrics/PerceivedComplexity
+                                       plate_join: :plate, tube_join: :tube, well_join: :well)
+      rec_ids = []
```
stevieing, line 43: personally I would use full names as there is no discernable difference from a syntactic view e.g record_ids, filtered_records. It removes ambiguity

app/models/concerns/source_identifier_filterable.rb
```diff
@@ -0,0 +1,73 @@
...
+#
+#   # Custom join associations
+#   filtered_records = MyModel.apply_source_identifier_filter(records, source_identifiers,
+#                                                             plate_join: :source_plate,
+#                                                             tube_join: :source_tube)
+#
+module SourceIdentifierFilterable
+  extend ActiveSupport::Concern
+
+  class_methods do # rubocop:disable Metrics/BlockLength
```
stevieing, line 40: I am not sure it is the right idea to ignore this. This is a great piece of work but if you split it into functions it will be a lot more readable. Happy to go through it with you.

seenanair: Ok. I will see I can break it down further. 👍

stevieing: I suspect Rubocop would have some ideas too but because of the Metrics/BlockLength it won't make any more suggestions.

app/models/concerns/source_identifier_filterable.rb
BenTopping, line 1: Should this be in shared, resources/v1/shared instead of as a model concern, similar to run_suitability?

BenTopping (changes requested):
Nice work.

I think we can also use this in resources/v1/ont/request_resource.rb for source_identifier.

app/models/concerns/source_identifier_filterable.rb
seenanair: I’ve broken the code into functions and modularized it using module functions. If these are changed to private class methods, RuboCop still raises concerns about the module size. Alternatively, wrapping these methods into another module and include it within this concern will allow  to remove 'block length' warning .

I’m not entirely convinced about the module size limits imposed by RuboCop  or the need to artificially modify the code just to satisfy it. While there is definitely value in splitting functions to improve readability and modularity, the imposed limit and the way RuboCop evaluates the structure feels bit unjustified.

stevieing: I understand your frustration and Rubocop can be dogmatic sometimes but looking at the result I find it far easier to understand what is going on. It is a question of maintainability. 

I think if we ignore Rubocop when we disagree with it we just end up with lots of stuff disabled which can defeat the purpose of it. I really struggle with long methods because I get overwhelmed and need to break it down. I would now feel far more confident modifying it because I know what is going on. The method names help me to understand. I hope that adds some context?

stevieing: Another thing this does for me is it helps me to identify places where we can make stuff more flexible. For example in this filter you have used plate, tube and well distinctions. I see this a lot in our code which leads me to believe that we need to do some work around labware. Having a big method can sometimes obfuscate this.

seenanair: Yes, absolutely. I fully agree with your point, and breaking down the structures absolutely made more sense. As you said, it is more readable, modular, and clear now. 
What I was referring to was the structure it seems to be approving of. It would have been equally good, or even better, to have the new methods added as private in-class methods, but RuboCop is not happy with that because of the size. This is the only point I am not convinced about.

> I understand your frustration and Rubocop can be dogmatic sometimes but looking at the result I find it far easier to understand what is going on. It is a question of maintainability.
> 
> I think if we ignore Rubocop when we disagree with it we just end up with lots of stuff disabled which can defeat the purpose of it. I really struggle with long methods because I get overwhelmed and need to break it down. I would now feel far more confident modifying it because I know what is going on. The method names help me to understand. I hope that adds some context?

stevieing: You are correct. It would have been better to have private in-class methods. Looking at what you have done I think there is a way we could reduce the size significantly by changing the structure. For example if we turned joins into a hash we could probably use map or reduce and build the filter in one go using Arel.

seenanair: I'm not entirely sure how to do it, may be we can do it by pairing on it at sometime later after finishing y24-149, if that's ok.

stevieing: Sure. I was going to have a crack but not quite sure either so pairing is a good idea.

stevieing (approved):
Only one comment.

Really good. Nice work. I especially like the tests for the shared.

We could think about shared specs for the resources but maybe when we use it next.

BenTopping (approved):

PR #1564, opened 2025-02-20
remove shapyr tables

sabrine33:


db/migrate/20250220115944_drop_shaphyr_tables.rb
```diff
@@ -0,0 +1,63 @@
+class DropShaphyrTables < ActiveRecord::Migration[7.2]
+  def up
+    drop_table :saphyr_chips, if_exists: true
```
BenTopping, line 3: Don't think this `if_exists` is needed.

db/migrate/20250220115944_drop_shaphyr_tables.rb
```diff
@@ -0,0 +1,63 @@
...
+    drop_table :saphyr_enzymes, if_exists: true
+    drop_table :saphyr_flowcells, if_exists: true
+    drop_table :saphyr_libraries, if_exists: true
+    drop_table :saphyr_requests, if_exists: true
+    drop_table :saphyr_runs, if_exists: true
+  end
+
+  def down
+    create_table "saphyr_chips", charset: "utf8mb3", force: :cascade do |t|
```
BenTopping, line 12: Think down can just be 
```
  def down
    fail ActiveRecord::IrreversibleMigration
  end
```
As this isn't really reversible because of the data loss.

BenTopping (changes requested):

db/migrate/20250220115944_drop_shaphyr_tables.rb
```diff
@@ -0,0 +1,63 @@
+class DropShaphyrTables < ActiveRecord::Migration[7.2]
```
BenTopping, line 1: DropSaphyrTables*

sabrine33: not sure what you mean here

stevieing: ShaphyrTables > SaphyrTables

stevieing (approved):
We will need to wait to hear back from Camille before merging in.
It also makes sense to merge this one first before the rebasecalling work which also has a migration.

PR #1587, opened 2025-03-19
Y25-169 - Add barcode attribute to PacBio Library and Pool resources

BenTopping:
Part of https://github.com/sanger/traction-ui/issues/2228

- Add barcode attribute to PacBio Library and Pool model via delegation to tube.
- Add barcode attribute to API resource for PacBio Library and Pool.

### DIFF
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -72,10 +72,12 @@ module V1
       #   @return [Float] the used volume of the library
       # @!attribute [rw] available_volume
       #   @return [Float] the available volume of the library
+      # @!attribute [r] barcode
+      #  @return [String] the barcode of the library (via tube)
       attributes :state, :volume, :concentration, :template_prep_kit_box_barcode,
                  :insert_size, :created_at, :deactivated_at, :source_identifier,
                  :pacbio_request_id, :tag_id, :primary_aliquot_attributes,
-                 :used_volume, :available_volume
+                 :used_volume, :available_volume, :barcode
 
       has_one :request, always_include_optional_linkage_data: true
       # If we don't specify the relation_name here, jsonapi-resources
--- a/app/resources/v1/pacbio/library_resource.rb
+++ b/app/resources/v1/pacbio/library_resource.rb
@@ -72,12 +72,14 @@ module V1
       #   @return [Float] the used volume of the library
       # @!attribute [rw] available_volume
       #   @return [Float] the available volume of the library
-      # @!attribute [r] barcode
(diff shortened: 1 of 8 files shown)


app/resources/v1/pacbio/library_resource.rb
```diff
@@ -72,10 +72,12 @@ class LibraryResource < JSONAPI::Resource
       #   @return [Float] the used volume of the library
       # @!attribute [rw] available_volume
       #   @return [Float] the available volume of the library
+      # @!attribute [r] barcode
```
stevieing, line 75: we have this as r but it is not marked as ready only. Is this an issue?

app/resources/v1/pacbio/pool_resource.rb
```diff
@@ -87,10 +87,12 @@ class PoolResource < JSONAPI::Resource
       #   @return [Float] the used volume of the pool
       # @!attribute [rw] available_volume
       #   @return [Float] the available volume of the pool
+      # @!attribute [r] barcode
+      #   @return [String] the barcode of the pool (via tube)
       attributes :volume, :concentration, :template_prep_kit_box_barcode,
                  :insert_size, :created_at, :updated_at,
                  :library_attributes, :used_aliquots_attributes, :primary_aliquot_attributes,
-                 :used_volume, :available_volume
+                 :used_volume, :available_volume, :barcode
```
stevieing, line 95: we have this as r but it is not marked as ready only. Is this an issue?

stevieing (changes requested):

app/resources/v1/pacbio/library_resource.rb
BenTopping: You will get a NoMethodError if you try and set this value as its being delegated to tube, but delegation does not allow you to set the values.

BenTopping: Ill override the creatable and updatable fields so you can't use barcode, that way we will get a better error.

app/resources/v1/pacbio/pool_resource.rb
BenTopping: Updated it [027bc54](https://github.com/sanger/traction-service/pull/1587/commits/027bc544038ef93648319046311cf8d9dfda4e98)

app/resources/v1/pacbio/library_resource.rb
BenTopping: Updated in [027bc54](https://github.com/sanger/traction-service/pull/1587/commits/027bc544038ef93648319046311cf8d9dfda4e98)

PR #1588, opened 2025-03-20
Y25-080 - rename comments to barcodes_and_concentrations in Pacbio run model and update related references

stevieing:
Closes #

- 

Previously we were inserting a space at the start of the barcodes and concentrations. I see no need to continue doing this but at the same time see no need to modify existing records.

### DIFF
--- a/app/resources/v1/pacbio/run_resource.rb
+++ b/app/resources/v1/pacbio/run_resource.rb
@@ -48,8 +48,7 @@ module V1
       #   @return [String] the creation time of the run
       # @!attribute [rw] state
       #   @return [String] the state of the run
-      # @!attribute [rw] comments
-      #   @return [String] the comments on the run
+
       # @!attribute [rw] pacbio_smrt_link_version_id
       #   @return [Integer] the ID of the PacBio SMRT Link version
       # @!attribute [rw] plates_attributes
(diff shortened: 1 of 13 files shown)


app/resources/v1/pacbio/run_resource.rb
```diff
@@ -59,10 +58,18 @@ class RunResource < JSONAPI::Resource
...
                  :pacbio_smrt_link_version_id, :plates_attributes,
                  :adaptive_loading, :sequencing_kit_box_barcodes
 
+      # @!attribute [r] barcodes_and_concentrations
+      #   @return [String] the barcodes and concentrations of the run
+      attribute :barcodes_and_concentrations, readonly: true
+
+      # @!attribute [r] comments [DEPRECATED]
+      #   @return [String] the comments on the run
+      attribute :comments, readonly: true
```
seenanair, line 71: Do we still need this comments attribute as we are replacing it with barcodes_and_concentrations?

spec/requests/v1/pacbio/runs_spec.rb
```diff
@@ -1282,7 +1268,6 @@
           run.reload
           expect(run.dna_control_complex_box_barcode).to eq 'Lxxxxx101717600123191_updated'
           expect(run.system_name).to eq 'Sequel I'
-          expect(run.comments).to eq 'A Run Comment updated'
```
seenanair, line 1285: Do we need to make a similar check for barcodes_and_concentrations?

app/resources/v1/pacbio/run_resource.rb
stevieing: I have left it in as a deprecation just in case there are any side effects.

seenanair (approved):

BenTopping (approved):

PR #1601, opened 2025-03-31
Upgrade Rails to 8.0 and update related gems

stevieing:
Closes #

- 


Gemfile
```diff
@@ -45,7 +45,7 @@ group :development, :test do
   gem 'webmock'
 end
 
-gem 'flipper', '~> 0.25.0'
```
BenTopping, line 48: Did you find out the reason this was locked? I think when I tried to unlock it in sequencescape it came up with data migrations because the way they stored the flags changed after version 1.

BenTopping (approved):
Looks good, appears a smoother transition than 7.2 and 7.1 👍

Gemfile
stevieing: No idea. I will try adding a flag to make sure it still works.

PR #1642, opened 2025-06-26
Y25-343 - Added support for asymmetric barcodes

### LINKED ISSUE #1623: Y25-343 - As Developers we need to add support for asymmetric barcodes so that long read can use them for Pacbio multiplexing
**User story**

As Developers we need to add support for asymmetric barcodes so that long read can use them for Pacbio multiplexing.

Currently Craig is doing this manually by adding the asymmetric tag in the adapter 2 field. This is a manual process and creates extra work 

Craig / Steve

Craig / Long Read

**Acceptance criteria**

- [x] add an optional oligo_reverse attribute which has the same properties as the oligo attribute
- [x] modify the rake task so that we can import asymmetric tag sets

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

borhara:
Closes #1623 

- Changes to schema and a migration for the optional `oligo_reverse` field
- Added `oligo_reverse` attribute to the tag resource
- Added a rake task with a parser for adding tag sets with asymmetric barcodes


borhara:
@stevieing Any suggestions for what the tag group should be called? I've used PacBio asymmetrical for now. And will there be more than one of those tag groups? Could be worth extracting the parsing somewhere else.

lib/tasks/asymmetric.txt
BenTopping, line 1: might be clearer to move this to a subdirectory so we can keep future task data in one place. Maybe _lib/tasks/data_. Also renaming to pacbio specific as we may get more asymmetric tagsets in the future.

lib/tasks/create_tags.rake
```diff
@@ -293,6 +294,44 @@ namespace :tags do
...
+      file_path = Rails.root.join('lib/tasks/asymmetric.txt')
+      lines = File.read(file_path).strip.split("\n")
+
+      parsed_tags = {}
+
+      lines.each_slice(2) do |header, sequence|
+        next unless header&.start_with?('>')
+
+        group_id, _, direction = header[1..].rpartition('_')
+        parsed_tags[group_id] ||= { group_id: group_id }
```
BenTopping, line 319: Im not sure how we deal with these tags so might be wrong but do we now have two group_ids? one forward and one back?

BenTopping (commented):
Good start. Couple questions/comments

- Needs a test in spec/models/tag_spec to check the new field validation.
- We need to update sample sheets I think. We have methods like same_barcodes_on_both_ends_of_sequence which i dont think will behave correctly with these tag sets and https://github.com/sanger/traction-service/blob/develop/app/exchanges/run_csv/pacbio_sample_sheet.rb#L94. I am not sure on the correct behaviour so might need to ask Craig.

lib/tasks/create_tags.rake
borhara: We have one group_id, two tags would be in the same group. But the format of the file seems to have an identifier at the end of the group to specify whether the barcode below is a forward or reverse barcode.

lib/tasks/asymmetric.txt
borhara: Agreed, thank you.

BenTopping (commented):
Nice looks good. Please could you update the lib/data txt file too with the new name.

PR #1644, opened 2025-06-27
improve resource to return sample info

wendyyang:
Closes #

- 

### DIFF
--- a/spec/resources/v1/sample_resource_spec.rb
+++ b/spec/resources/v1/sample_resource_spec.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+# spec/requests/v1/samples_spec.rb
+require 'rails_helper'
+
+RSpec.describe 'V1::Samples API', type: :request do
+  describe 'GET /v1/samples/:id' do
+    let(:sample) { create(:sample, name: 'Sample X', created_at: Time.zone.local(2024, 6, 1)) }
+
+    let(:pacbio_requestable) do
+      create(:pacbio_request, external_study_id: '3b1cf0ac-4079-11f0-805f-e2df7c04b5f6', cost_code: 'COST456')
+    end
+
+    before do
+      create(:request,
+             sample: sample,
+             requestable_type: 'Pacbio::Request',
+             requestable: pacbio_requestable)
+    end
+
+    it 'resource include sample name' do
+      get "/v1/samples/#{sample.id}"
+
+      expect(response).to have_http_status(:ok)
+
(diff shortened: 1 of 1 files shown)


db/migrate/20250627145127_change_flipper_gates_value_to_text.rb
BenTopping, line 1: Where did this change come from? Is it an update to flipper gem?

app/resources/v1/sample_resource.rb
```diff
@@ -43,6 +43,19 @@ class SampleResource < JSONAPI::Resource
     attributes :sanger_sample_id, :supplier_name, :taxon_id, :donor_id, :country_of_origin,
                :accession_number, :date_of_sample_collection
 
+    filter :name
+    attribute :pacbio_requests
```
BenTopping, line 47: I think we should be able to do pacbio_requests as a relationship through request as opposed to an attribute. Do we need ont_requests too?

BenTopping (commented):
I think we can get this information more easily from the pacbio_request resource  [here](https://github.com/sanger/traction-service/blob/develop/app/resources/v1/pacbio/request_resource.rb). I think this story involves updating the filters in that resource so we can include samples there and filter by multiple barcodes / sample names but not entirely sure. cc @stevieing. I am also unsure on whether we need to support ont requests too?

db/migrate/20250627145127_change_flipper_gates_value_to_text.rb
wendyyang: I think I merge develop into my branch and got this change.

wendyyang:
> I think we can get this information more easily from the pacbio_request resource [here](https://github.com/sanger/traction-service/blob/develop/app/resources/v1/pacbio/request_resource.rb). I think this story involves updating the filters in that resource so we can include samples there and filter by multiple barcodes / sample names but not entirely sure. cc @stevieing. I am also unsure on whether we need to support ont requests too?

What does traction UI side expect service to return?

db/migrate/20250627145127_change_flipper_gates_value_to_text.rb
stevieing: This is not in develop as far as I can see?

app/resources/v1/sample_resource.rb
stevieing: Should we do this the other way round? Should we use requests and pull in the sample information? That way we could have a different report for Pacbio and ONT (if required) to reduce complexity especially since they are so different?

PR #1651, opened 2025-07-02
Y25-344 - Adapter 2 includes oligo_reverse attribute on sample sheet

### LINKED ISSUE #1624: As long read Craig I would like the sample sheet to have adapter 2 for tag sets that have asymmetric barcodes
**User story**

As long read Craig I would like the sample sheet to have adapter 2 for tag sets that have asymmetric barcodes

Currently Craig is doing this manually by adding the asymmetric tag in the adapter 2 field. This is a manual process and creates extra work

Craig / Steve

Craig / Long Read

**Acceptance criteria**

- [ ] when I print a sample sheet with an asymmetric tag set it will add the oligo_reverse to the adapter 2 field

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

borhara:
Closes #1624 

- If the `oligo_reverse` attribute is present on the linked tag, it will be used as the value for the adapter 2 field on the sample sheet

### DIFF
--- a/app/exchanges/run_csv/pacbio_sample_sheet.rb
+++ b/app/exchanges/run_csv/pacbio_sample_sheet.rb
@@ -91,7 +91,7 @@ module RunCsv
         bio_sample_name(sample),
         well.plate_well_position,
         sample.adapter, # left adapter
-        sample.adapter  # right adapter
+        sample.library&.tag&.oligo_reverse.presence || sample.adapter # right adapter
       ]
     end
 


app/exchanges/run_csv/pacbio_sample_sheet.rb
```diff
@@ -91,7 +91,7 @@ def sample_data(well, sample)
         bio_sample_name(sample),
         well.plate_well_position,
         sample.adapter, # left adapter
-        sample.adapter  # right adapter
+        sample.library&.tag&.oligo_reverse.presence || sample.adapter # right adapter
```
BenTopping, line 94: This will return the reverse oligo instead of the group_id. Just wanted to check this is intended as the current adapter returns the group_id.

I think this logic should be moved to a method like above e.g. sample.right_adapter. 

Do we need to do .library? Sample in this case should be an aliquot which might already have the tag. If we do just library this may not work correctly for pools.

BenTopping (commented):
Looks ok, small comments. Have you tested a sample sheet using reverse tags on the smrtlink test instance?

app/exchanges/run_csv/pacbio_sample_sheet.rb
```diff
@@ -87,11 +87,12 @@ def smrt_cell_settings
     end
 
     def sample_data(well, sample)
+      adapter_suffix = sample.tag&.oligo_reverse.present? ? %w[_L _R] : ['', '']
```
stevieing, line 90: I am just wondering if it would be better to have methods in the tag model adapter_1 and adapter_2? They are then more easily testable and maintanable.

borhara:
> I presume the adapter_1 and adapter_2 methods are tested somewhere?

Not yet, although the other adapter method isn't directly testing anywhere, just in the sample sheet spec.

PR #1680, opened 2025-08-19
Y25-522 - Allow import of mulitple compound samples with the same supplier name

### LINKED ISSUE #1679: [BUG] As a user (Mary S) I would like to be able to import compound samples if they have the same supplier name
**Describe the bug**
[BUG] As a user (Mary S) I would like to be able to import compound samples if they have the same supplier name

**RT Ticket Number**
If applicable

**To Reproduce**
1. Import a compound sample tube via Kinnex reception.
2. Import a different compound sample tube with the same supplier name as the first via Kinnex reception.
3. Get name error.

**Expected behaviour**
Should allow import

BenTopping:
Closes #1679 

- Compound samples use existing sample if already exists, otherwise they create a new one.

###
I have tested this with some of the prod data the users had that caused this bug and import fine.

### DIFF
--- a/app/models/reception/resource_factory.rb
+++ b/app/models/reception/resource_factory.rb
@@ -170,6 +170,9 @@ class Reception
       supplier_name = first_sample[:supplier_name]
       species = first_sample[:species]
       compound_sample = create_compound_sample(supplier_name, species)
+      # If its an existing sample if will not have initialised the component sample uuids
+      # So we check here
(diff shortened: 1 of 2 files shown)


BenTopping:
My main concern with this is whether the warehouse will accept multiple compound samples with the same external_id.

stevieing (approved):

PR #1782, opened 2026-02-13
feat(multi-pools): updates positions to be integers instead of coordinates

BenTopping:

- Updates multi pool seed data, docs and testing references to use integer for positions

#### Additional notes
This comes from [a conversation here: ](https://github.com/sanger/traction-ui/pull/2596). Pools are group by numbers no coordinates.

### DIFF
--- a/db/migrate/20260213150155_modify_multi_pool_position_position_to_int.rb
+++ b/db/migrate/20260213150155_modify_multi_pool_position_position_to_int.rb
@@ -1,0 +1,5 @@
+class ModifyMultiPoolPositionPositionToInt < ActiveRecord::Migration[8.1]
+  def change
+    change_column :multi_pool_positions, :position, :integer
+  end
+end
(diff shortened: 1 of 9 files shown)


db/schema.rb
BenTopping, line 1: This is caused by rails 8.1 https://rubyonrails.org/2025/1/10/this-week-in-rails.

db/migrate/20260213150155_modify_multi_pool_position_position_to_int.rb
```diff
@@ -0,0 +1,5 @@
+class ModifyMultiPoolPositionPositionToInt < ActiveRecord::Migration[8.1]
+  def change
+    change_column :multi_pool_positions, :position, :integer
```
yoldas, line 3: note: This fixes the data type to integer (it could stay as string and converted in the code), but we can drive coordinates (A1 ... H12) from integers in case we need to use string positions in the UI.

yoldas (approved):

PR #1799, opened 2026-03-13
Y26-046 - Keep pacbio library requests and used_aliquot sources in sync

### LINKED ISSUE #1788: Y26-046 - As a developer (Ben) I would like to cleanup the relationship between pacbio requests and libraries
**Describe the Housekeeping**
Currently a pacbio_request can relate to the pacbio library in two ways:

1. Through used_aliquots. A pacbio library 'uses' a request aliquot so we can find the relationship between the two this way. A pacbio library should only ever have one used_aliquot and this is always a request currently.
2. Through pacbio_request_id attribute (Foreign key).

We should consolidate the two ways (remove pacbio_request_id) or add logic to ensure they are always kept in sync. An example of the issues/complexity this causes can be seen here https://github.com/sanger/traction-service/pull/1787.

If we decide to remove pacbio_request_id FK relationship we should add a script to ensure the used_aliquots and requests are in sync to ensure we don't break any data. We should also check this does not break any UI behaviour / logic.

BenTopping:
Closes #1788 

- Sets source in pacbio library used_aliquots during after_update hook.

This is the PR for the research story associated above. A brief explanation is available as a comment there.

### DIFF
--- a/app/models/pacbio/library.rb
+++ b/app/models/pacbio/library.rb
@@ -80,7 +80,10 @@ module Pacbio
           concentration: primary_aliquot.concentration,
           template_prep_kit_box_barcode: primary_aliquot.template_prep_kit_box_barcode,
           insert_size: primary_aliquot.insert_size,
-          tag: primary_aliquot.tag
+          tag: primary_aliquot.tag,
+          # If we change the request (e.g. sample swap) we also need to update the source of the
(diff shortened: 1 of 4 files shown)


StephenHulme (approved):
Nice linked explanation - thanks!

Issue #1735, opened 2025-11-05
Y25-661 - As a developer (Ben) I would like to integrate the Okta login system.


BenTopping:
**User story**

As a developer (Ben) I would like to integrate the Okta login system. Following on from the traction-ui story the service should validate the provided access token. Following the [client credentials flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow) the service should validate the token provided by the UI.

API keys and programmatic access will be handled by a separate story so the API should remain generally accessible but we should still be authenticated the tokens for when the API keys are implemented.

**Who are the primary contacts for this story**

Ben T

**Who is the nominated tester for UAT**

Developer

**Acceptance criteria**

Consider if these features can be _featured flagged_ to decouple testing and deployment.
To be considered successful the solution must allow:
- [ ] All API endpoints validates Traction ui okta token
- [ ] ...

**Dependencies**

This story is blocked by the following dependencies:
- Dependent on UI story https://github.com/sanger/traction-ui/issues/2534

**Additional context**
This is an outcome of https://github.com/sanger/traction-ui/issues/2430

BenTopping:
Moved to triage while https://github.com/sanger/traction-ui/issues/2534 is worked on as scope of work may change.

yoldas:
### Notes:

Screenshot: Request to JKWS endpoint showing the kid for the public key. Traction Service token validation picks the right key using the kid from the token payload, and validates the signing signature among other validations.

<img width="480" height="384" alt="Image" src="https://github.com/user-attachments/assets/28b28a3e-f973-435e-8836-7783b633cee2" />

PR #1803, opened 2026-03-20
Y25-662 - [PR] As a developer (Ben) I would like to add API keys to enable programmatic access to traction-service

### LINKED ISSUE #1736: Y25-662 - As a developer (Ben) I would like to add API keys to enable programmatic access to traction-service
**User story**

As a developer (Ben) I would like to add API keys to enable programmatic access to traction-ui. This includes adding a new table to store hashed API keys and validating them via the API. If user authentication is already implemented via https://github.com/sanger/traction-service/issues/1735 then the API can be moved completely behind auth and no routes ~API key users should only be able to get from the API, not post.~

Ben T

Developer

**Acceptance criteria**

- [x] A new table is added to store api keys and includes the name of the user.
- [x] API keys are validated on get requests to the API.
- [x] New api keys are easily rotated or created.
- [ ] Existing programmatic users are notified and given an API key. This may require identifying and contacting users (TOL , NPG etc) tol-lab-share? [see additional context]

**Additional context**
Rollout strategy includes enabling the following feature flags:

1. **25_662_enable_request_authentication**: When enabled, this will activate the new authenticator module
2. **y25_662_reject_unauthenticated_requests**: When enabled, this will reject any unauthenticated requests

Until the second enabled, thanks to the logging of unauthenticated requests, we will be able to identify the programmatic users  that require API Key access.

This is an outcome of https://github.com/sanger/traction-ui/issues/2430

yoldas:
Closes #

Adds API key support for programmatic access to Traction service

- Added ApiApplication and ApiKey models.
- Implemented API key issuance using one-time plaintext tokens with persisted SHA-256 key_digest only.
- Implemented lifecycle rotation in ApiApplication#rotate_api_key!:
  - active -> grace_period
  - grace_period -> expired
  - issue new active key

- Added Authenticator concern flow for:
  - Bearer token auth path
  - API key auth via X-Traction-Client-Id
  - GET-only enforcement for API-key-authenticated requests
  - grace-period warning header
  - expires_at enforcement in auth checks [We are allowing non-expiring keys for now]

- Added migration and schema updates for:
  - api_applications
  - api_keys (key_digest, status, expires_at, last_used_at)

- Added rake task:
  - api_keys:rotate[api_application_name,expires_at] for single-app key rotation by name.

- Added factories and full test coverage for models, authenticator flows, and rake task behaviour.

Notes:
- ~API keys are treated as read-only credentials (GET requests only)~ [post using API Key should be allowed]
- A key can be rejected either by lifecycle status (expired) or by expires_at having passed.
- Grace-period keys remain valid but return a deprecation warning header.
- Marked the places for Y25-661 (OKTA) https://github.com/sanger/traction-service/issues/1735

### DIFF
--- a/app/controllers/concerns/authenticator.rb
+++ b/app/controllers/concerns/authenticator.rb
@@ -129,7 +129,7 @@ module Authenticator
 
     return if request.get?
 
-    render_forbidden('API keys are read-only')
+    render_forbidden('API keys are for read-only access')
   end
 
   # Handles requests with no valid credentials.
@@ -178,7 +178,6 @@ module Authenticator
     key.expires_at.present? && key.expires_at.past?
   end
 
-  # Response helpers
   # Renders a 401 Unauthorized response.
   #
   # @param message [String] error message to include in response
--- a/app/controllers/concerns/authenticator.rb
+++ b/app/controllers/concerns/authenticator.rb
@@ -6,7 +6,6 @@
 # 1. Bearer tokens (JWT/OAuth2, for future Identity Provider integration)
 # 2. API keys (via X-Traction-Client-Id header, for programmatic access)
 #
-# Enforces API key permissions: keys are read-only (GET requests only).
 # Key usage is logged, and grace-period keys trigger deprecation warnings.
 #
 # @example Include in a controller
@@ -41,7 +40,6 @@ module Authenticator
       authenticate_bearer!
     elsif api_key_present?
       authenticate_api_key!
-      enforce_api_key_permissions!
     else
       handle_unauthenticated!
     end
@@ -121,17 +119,6 @@ module Authenticator
     @current_api_client = key.api_application
   end
 
-  # Enforces that API key-authenticated requests are GET-only (read-only).
-  #
-  # @return [void]
-  def enforce_api_key_permissions!
-    return unless @current_auth == :api_key
-
-    return if request.get?
-
-    render_forbidden('API keys are for read-only access')
-  end
-
   # Handles requests with no valid credentials.
   #
   # @return [void]
@@ -193,13 +180,5 @@ module Authenticator
   # @return [void]
   def render_unauthorized(message)
     render json: { error: message }, status: :unauthorized
-  end
-
-  # Renders a 403 Forbidden response.
-  #
-  # @param message [String] error message to include in response
-  # @return [void]
-  def render_forbidden(message)
-    render json: { error: message }, status: :forbidden
   end
 end
--- a/app/controllers/concerns/authenticator.rb
+++ b/app/controllers/concerns/authenticator.rb
@@ -9,7 +9,8 @@
 module Authenticator
   extend ActiveSupport::Concern
 
-  API_KEY_HEADER = 'X-Traction-Client-Id' # Using headers instead of env, HTTP_X_TRACTION_CLIENT_ID in Rails.
+  # Using headers instead of env, HTTP_X_TRACTION_CLIENT_ID in Rails.
+  API_KEY_HEADER = 'X-Traction-Client-Id'
 
   included do
     before_action :authenticate_request! # preceed all controller actions
--- a/app/controllers/concerns/authenticator.rb
+++ b/app/controllers/concerns/authenticator.rb
@@ -20,6 +20,8 @@ module Authenticator
 
   # Main authentication entry point
   def authenticate_request!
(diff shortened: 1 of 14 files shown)


app/controllers/concerns/authenticator.rb
```diff
@@ -0,0 +1,196 @@
...
+    return if request.get?
+
+    render_forbidden('API keys are for read-only access')
+  end
+
+  # Handles requests with no valid credentials.
+  #
+  # @return [void]
+  def handle_unauthenticated!
+    render_unauthorized('Authentication required')
```
BenTopping, line 135: For first implementation could we put a logger here instead so we know which users are accessing unauthenticated and can distribute keys accordingly.

spec/controllers/concerns/authenticator_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    end
+
+    def create
+      render plain: 'created'
+    end
+  end
+
+  before do
+    routes.draw do
+      get 'index' => 'anonymous#index'
```
BenTopping, line 29: How does this `anonymous` bit work?

spec/controllers/concerns/authenticator_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  it 'authenticates a valid API key for GET requests' do
+    allow(Flipper).to receive(:enabled?).with(feature).and_return(true)
+    issued = api_application.rotate_api_key!
+
+    request.headers['X-Traction-Client-Id'] = issued[:plaintext_key]
+    get :index
+
+    expect(response).to have_http_status(:ok)
+    expect(issued[:api_key].reload.last_used_at).not_to be_nil
```
BenTopping, line 59: Could also check a log is output

spec/controllers/concerns/authenticator_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+  it 'rejects malformed API keys' do
+    allow(Flipper).to receive(:enabled?).with(feature).and_return(true)
+
+    request.headers['X-Traction-Client-Id'] = 'bad-format-key'
+    get :index
+
+    expect(response).to have_http_status(:unauthorized)
+  end
+
+  it 'forbids non-GET requests authenticated by API key' do
```
BenTopping: Thinking about this tol-lab-share will be able to post to traction reception over the API using an API key. I think we can get rid of this non-get criteria for initial implementation. Apologies!

spec/lib/tasks/api_keys.rake_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+RSpec.describe 'RakeTasks' do
+  describe 'api_keys:rotate' do
+    before do
+      Rake::Task['api_keys:rotate'].reenable
+    end
+
+    it 'rotates an API key for a single application by name' do
+      api_application = create(:api_application)
+
```
BenTopping, line 17: Check active key moves to grace period?

app/controllers/concerns/authenticator.rb
BenTopping: Maybe another feature flag?

BenTopping (commented):
Nice work, really good.

I think we should remove the criteria for just 'GET' requests via API keys on reflection as tol-lab-share will need to post via api key.

spec/controllers/concerns/authenticator_spec.rb
yoldas: controller(ApplicationController) takes a block to define methods on a temporary controller called AnonymousController. routes.draw takes a block to define routes for the anonymous controller to test requests.

spec/controllers/concerns/authenticator_spec.rb
yoldas: Removed `enforce_api_key_permissions!` . We may consider implementing privileges for API apps though.

spec/lib/tasks/api_keys.rake_spec.rb
yoldas: It is tested in `spec/models/api_application_spec.rb` .

spec/controllers/concerns/authenticator_spec.rb
yoldas: I have improved Authenticator#log_key_usage to log active or grace-period keys and added assertions in the tests now.

BenTopping (approved):

PR #1806, opened 2026-03-23
Y26-091 - Multi pool volume check

### LINKED ISSUE #1796: Y26-091 - [BUG] As a developer (Ben) I would like to ensure volume is validated across pools in a multi pool
**Describe the bug**
As a developer (Ben) I would like to ensure volume is validated across pools in a multi pool. Currently this does not occur so if a library is used in multiple positions within one multi pool and the total sum of volume exceeds the libraries available volume it will not flag as an error. This can lead to using more volume than is possible and reaching negative volume for a library.

This in theory shouldn't be an issue in the lab as you shouldnt be able to use more volume than available but we want to check against it to prevent and flag accidental mistakes. We can leave this as jusst a server side validation as it will be complex as a UI guard and rarely used.

**To Reproduce**
1. Go to flexible pooling page
2. Scan a library into multiple positions in a multi pool and use the full volume in each position
3. See no error occurs.

**Expected behaviour**
An error occurs saying the available volume has been exceeded for the given library.

**Additional context**
Suggested fix:

Add a validation to multi_pool model that checks pipeline is pacbio and if any sources have been used across multiple positions. If so check the sum of volume for the sources is not exceeding the available volume.

BenTopping:
Closes #1796 

- Moves aliquot volume checks to aliquot model validation
    - This was ultimately not required but it is a more logical place for the checks. The previous checks also were not working as intended, they did not work on creation/update via the API due to the race conditions mentioned. 
- Adds multi_pool check to ensure pool's library volumes arent exceeded.
    - Unfortunately this is a bit ugly, it would be nicer for this to be handled with the rest of the volume checks. However this is the only case where multiple derived_aliquots can be created simultaneously for the same source. E.g. two pools use the same library and are created at the same time. This race condition bypasses the current volume checks as they are not aware of the other pool being created. To handle this we check at the multi pool level that if any libraries have been used multiple times their combined volume doesnt exceed the libraries available volume.


spec/models/concerns/aliquotable_spec.rb
```diff
@@ -24,6 +24,8 @@
     it 'returns the derived aliquots' do
       pacbio_pool = create(:pacbio_pool)
       aliquots = create_list(:aliquot, 5, aliquot_type: :derived, source: pacbio_pool)
+      pacbio_pool.reload
```
BenTopping, line 27: These reloads are caused by polymorphic relationships needing reloading. I think these were loaded automatically previously when the libraries/pools handled the volume check logic as they loaded the aliquots to check.

app/models/concerns/aliquotable.rb
```diff
@@ -57,49 +57,5 @@ def available_volume
     def available_volume_sufficient?
       available_volume >= 0
     end
-
```
dasunpubudumal, line 60: Good job taking them away from concerns. I never liked them. Turns out, there are other people [that do not like them](https://news.ycombinator.com/item?id=24512083). It reminds me too much of Java's Reflection API. They don't really behave the same way, but they are both meta-programming. It can also confuse language servers.

app/models/pacbio/library.rb
```diff
@@ -49,11 +49,6 @@ class Library < ApplicationRecord
 
     after_create :create_used_aliquot
 
-    # Before updating the record, the `primary_aliquot_volume_sufficient` method is called.
-    # This is a callback that checks if the primary_aliquot volume has been changed and
-    # if the new volume is greater than the used volume.
-    before_update :primary_aliquot_volume_sufficient
```
dasunpubudumal, line 55: Is the function body still there in the source code? If not used, maybe add some `@deprecated` annotation to remove it in a subsequent release? (I don't know the best way to do it in Ruby).

app/models/aliquot.rb
```diff
@@ -37,6 +37,50 @@ class Aliquot < ApplicationRecord
...
+
+  # Checks the derived aliquot does not exceed the available volume of the source
+  # See models/multi_pool.rb sufficient_library_used_volume? for additional edge case checks
+  def check_available_parent_volume
+    # Some sources may not have a primary aliquot, e.g. a Pacbio::Request, so we only want to check
+    # the volume if there is a primary aliquot to check against
+    return unless aliquot_type == 'derived' && source&.primary_aliquot&.volume
+
+    # For persisted records, add back the previous volume to avoid double subtraction
+    previous_volume = persisted? ? volume_was : 0
```
dasunpubudumal, line 51: Where is `volume_was` coming from 🤔

app/models/aliquot.rb
```diff
@@ -37,6 +37,50 @@ class Aliquot < ApplicationRecord
   validates :volume, :concentration, :insert_size,
             numericality: { greater_than_or_equal_to: 0, allow_nil: true }
 
+  validate :check_available_parent_volume
```
dasunpubudumal, line 40: It looks like the two validations do not depend on each other so this is something very _very_ minor: which order are these validations evaluated? What's running first?

Given that `primary_aliquot_volume_sufficient` method has less instructions, it probably needs to run before as there's no use of evaluating `check_available_parent_volume` if the former fails; saves compute. Not a big deal anyway. It all depends on how Rails unrolls all these functions to instructions.

Image

app/models/multi_pool.rb
```diff
@@ -48,4 +49,37 @@ def unique_pool_positions?
...
+  # all pools in the multi pool.
+  def library_source_volume_map
+    library_source_volume_map = {}
+    multi_pool_positions.each do |position|
+      next unless position.pacbio_pool
+
+      position.pacbio_pool.used_aliquots.each do |aliquot|
+        next unless aliquot.source_type == 'Pacbio::Library'
+
+        library_source_volume_map[aliquot.source] ||= 0
```
dasunpubudumal: Can we do [`Hash#merge!`](https://apidock.com/ruby/Hash/merge!) here?

```ruby
library_source_volume_map.merge!(aliquot.source => aliquot.volume) { |_, old, new| old + new }
```

Image

dasunpubudumal (commented):
I think the logic is okay. Just some minor comments.

And I wasn't clear on where [`volume_was` symbol](https://github.com/sanger/traction-service/pull/1806/changes#diff-51d2fbcb09d9d89ef0c74e2bea88c7527fdb1e10fd47b141b58118f6ca052ee1R51) was coming from 🤔

app/models/concerns/aliquotable.rb
BenTopping: Thanks. Yeah they can be good but I think this case was the wrong use case because its relying on specific classes which is against the point of concerns.

app/models/pacbio/library.rb
BenTopping: Yes I have move the validation to the aliquot class. See https://github.com/sanger/traction-service/pull/1806/changes#diff-51d2fbcb09d9d89ef0c74e2bea88c7527fdb1e10fd47b141b58118f6ca052ee1R76

app/models/aliquot.rb
BenTopping: Its rails magic ✨ 
It gets the current value from the database before the current commit. https://api.rubyonrails.org/v5.2/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-attribute_in_database

app/models/aliquot.rb
BenTopping: Yes good point. Ill take a look into it. I think `primary_aliquot_volume_sufficient` should run first as thats a more fundamental check and has a higher chance of erroring.

app/models/aliquot.rb
dasunpubudumal: Oh god. The module is called `ActiveModel::Dirty`. The module is aptly named.

app/models/aliquot.rb
BenTopping: Learnt something new looking into this in that the declaration order in the model affects the order the validations are run.

It also looks like all validations are run _anyway_ even if an earlier validation fails so it wouldn't prevent the other one running unless we made the validation conditional (e.g. if no errors yet then run). While we could do this I think its easier to leave as is. These are low compute functions with early guard clauses to prevent excess running.

dasunpubudumal: It is weird for Rails to run all the validations. No wonder why that framework is slow.

app/models/multi_pool.rb
BenTopping: update in: [a4e3628](https://github.com/sanger/traction-service/pull/1806/commits/a4e3628a633204f0e8d01c74c4f3f87973c60e45)

dasunpubudumal (approved):

PR #1835, opened 2026-05-11
feat(auth): makes get requests permissive

BenTopping:

- Makes all get requests permissive

This was decided because traction-ui retrieves sample sheet's via get requests and Limber polls traction for tube creation. This may change in the future but it is fine for initial implementation.

### DIFF
--- a/app/controllers/concerns/authenticator.rb
+++ b/app/controllers/concerns/authenticator.rb
@@ -33,8 +33,14 @@ module Authenticator
   # 2. API key (if X-Traction-Client-Id header present)
   # 3. Denied (if neither present)
   #
+  # GET requests are allowed without authentication.
+  # This can be changed in the future but it was decided because of known use cases:
+  # - Limber check for tube creation
+  # - Traction-ui sample sheet generation link
+  #
   # @return [void]
   def authenticate_request!
+    return if request.get?
     return unless Flipper.enabled?(:y25_662_enable_request_authentication)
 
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/authenticator.rb
```diff
@@ -33,8 +33,14 @@ module Authenticator
...
   # 3. Denied (if neither present)
   #
+  # GET requests are allowed without authentication.
+  # This can be changed in the future but it was decided because of known use cases:
+  # - Limber check for tube creation
+  # - Traction-ui sample sheet generation link
+  #
   # @return [void]
   def authenticate_request!
+    return if request.get?
```
yoldas: I think putting it here short-circuits too early. It practically disables security checks for any GET requests. The purpose of this method is dispatching the request for three authentication methods: bearer_token, api_key, and unauthenticated. Can it go into `handle_unauthenticated!` ?

BenTopping: Okay happy to move it. I put it there to prevent unnecessary checks we know are going to fail `bearer_token_present?` and `api_key_present?`.

BenTopping: Updated in [1c32aff](https://github.com/sanger/traction-service/pull/1835/commits/1c32affc98a91dd267da48fdd05e8a2e2732bb99)

yoldas (approved):

PR #1868, opened 2026-07-03
Y26-179 - Merge smrtbell and pacbio96v3 tag sets.

### LINKED ISSUE #1849: Y26-179 - Disable Pacbio_96_barcode_plate_v3 and Transition to SMRTbell_Barcoded_Adapter_Plates_ABCD
**Context**
The current Pacbio_96_barcode_plate_v3 tag set is still available within the system, despite the introduction of the newer SMRTbell_Barcoded_Adapter_Plates_ABCD tag set. Continuing to use the older tag set limits operational capability and creates potential complications for future sample pooling.

The SMRTbell_Barcoded_Adapter_Plates_ABCD tag set provides approximately 300% more barcodes, significantly increasing sequencing flexibility and improving pooling capacity. However, to fully realise these benefits and avoid inconsistencies between legacy and new workflows, users need to be transitioned away from the old tag set.

This transition requires two coordinated activities:
- Disabling the Pacbio_96_barcode_plate_v3 tag set to prevent further use.
- Updating historical and existing samples currently assigned to Pacbio_96_barcode_plate_v3 so they instead reference SMRTbell_Barcoded_Adapter_Plates_ABCD.

Without this change, users may continue selecting outdated barcode plates, limiting scalability and increasing the risk of pooling-related issues.

**User story**
As a sequencing laboratory user, I want the Pacbio_96_barcode_plate_v3 tag set to be disabled and existing samples transitioned to SMRTbell_Barcoded_Adapter_Plates_ABCD, So that I can use the expanded barcode capacity, maintain consistency across workflows, and avoid future pooling issues.

James W

James W

**Acceptance criteria**
- [ ] The Pacbio_96_barcode_plate_v3 tag set can no longer be selected or assigned to new samples or workflows.
- [ ] Samples currently associated with Pacbio_96_barcode_plate_v3 are updated to use SMRTbell_Barcoded_Adapter_Plates_ABCD without loss of sample data or traceability.

**Dependencies**

- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

**Deployment notes**

BenTopping:
Closes #1849 

- Adds single_use lib rake folder for one off data migration scripts.
- Adds script to merge pacbio smrtbell tag sets.

#### Additional notes
The main reason for swapping the tags between the tag_sets is to prevent migrating a large amount of data on the models that use the tags e.g Pacbio::Library and Aliquot. If we didnt swap the tags and instead changed the underlying data to point to the equivalent tags we would touch ~20000 records.

Y26-179 - Traction tag set merge (1)


lib/tasks/create_tags.rake
```diff
@@ -452,7 +452,7 @@ namespace :tags do
     task pacbio_96_barcode_plate_v3: :environment do
       puts '-> Creating Pacbio_96_barcode_plate_v3 tag set and tags'
       set = TagSet.pacbio_pipeline
-                  .find_or_create_by!(name: 'Pacbio_96_barcode_plate_v3', uuid: '7a7f33e6-4912-4505-0d1e-3ceef7c93695')
+                  .find_or_create_by!(name: 'Pacbio_96_barcode_plate_v3', uuid: '7a7f33e6-4912-4505-0d1e-3ceef7c93695', active: false)
```
yoldas: This should use another approach because finding involves all three fields otherwise.
find_or_initialize by the first two fields, and then update .
or
find_or_create_by and then update.

BenTopping: AH yes good spot! Thanks.

BenTopping: Fixed in [5440c6e](https://github.com/sanger/traction-service/pull/1868/commits/5440c6e8e0ea3439befac2b9fdf5e9b024612729). Turns out we made this mistake in a few places. Good catch 👏

lib/tasks/single_use/2026_07_merge_pacbio_smrtbell_tag_sets.rake
```diff
@@ -0,0 +1,147 @@
...
+#   equivalent SMRTbell_Barcoded_Adapter_Plates_ABCD tags.
+#   - Note: Small migration ~300 records to update the libraries and request pool used aliquots to
+#     use the new tags. Tag group_id/oligo will remain the same so this is a simple update of the
+#     tag_id in the database.
+# - Deactivate the old Pacbio_96_barcode_plate_v3 tag set.
+#
+namespace :single_use do # rubocop:disable Metrics/BlockLength
+  # Check that the required tag sets exist and return them
+  def check_tag_sets_exist
+    p96v3_tag_set = TagSet.find_by(name: 'Pacbio_96_barcode_plate_v3')
```
BenTopping: check its active

lib/tasks/create_tags.rake
```diff
@@ -340,8 +343,8 @@ namespace :tags do
       puts '-> Creating Nextera UD tag set and tags'
       set = TagSet.pacbio_pipeline
                   .create_with(sample_sheet_behaviour: :hidden)
-                  .find_or_create_by!(name: 'Nextera_UD_Index_PlateA', uuid: 'dd87fafe-1432-157f-fd16-b1a8bae9e6a4', active: false)
-
+                  .find_or_create_by!(name: 'Nextera_UD_Index_PlateA', uuid: 'dd87fafe-1432-157f-fd16-b1a8bae9e6a4')
+      set.update!(active: false) if set.active?
```
StephenHulme, line 347: Are these all production UUIDs? Is it worth just referring to the names, or are the UUIDs the same in UAT too?

StephenHulme (approved):
Looks good, just one question.

lib/tasks/create_tags.rake
BenTopping: Yes UUID's should be the same across all envs, they link to production SMRTLink UUIDs so when we create them we have to make sure the UUIDs match.

yoldas (approved):

PR #1870, opened 2026-07-10
Remove stale feature flags

BenTopping:
Removes stale feature flags from the code base

- y25_662_enable_request_authentication [1ba6439](https://github.com/sanger/traction-service/pull/1870/commits/1ba64392387809c69c6d7fcc627b0cdb1e01d88a)
- y25_662_reject_unauthenticated_requests [85cd964](https://github.com/sanger/traction-service/pull/1870/commits/85cd964a40bcbd93d86df04981285f56d8f18bab)
    - This was slightly tricker as it means now that all of our non-GET specs require an API key to get past auth hence the mass changes from json_api_headers to -> auth_json_api_headers. This felt like a better solution that stubbing the auth module.

### DIFF
--- a/spec/support/request_helpers.rb
+++ b/spec/support/request_helpers.rb
@@ -8,6 +8,17 @@ module RequestHelpers
     }
   end
 
+  def auth_headers(api_application = create(:api_application))
+    issued = api_application.rotate_api_key!
+    {
+      'X-Traction-Client-Id' => issued[:plaintext_key]
+    }
+  end
+
+  def auth_json_api_headers(api_application = create(:api_application))
+    json_api_headers.merge(auth_headers(api_application))
+  end
+
   #
   # Returns a hash representing the decoded json response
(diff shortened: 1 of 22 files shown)


spec/support/request_helpers.rb
```diff
@@ -8,6 +8,17 @@ def json_api_headers
...
   end
 
+  def auth_headers(api_application = create(:api_application))
+    issued = api_application.rotate_api_key!
+    {
+      'X-Traction-Client-Id' => issued[:plaintext_key]
+    }
+  end
+
+  def auth_json_api_headers(api_application = create(:api_application))
```
yoldas, line 18: note: rotates the key at each call but it is good for test isolation.

yoldas (approved):
I appreciate that updating the tests after removing the feature flags was not straightforward. It now has a more secure authorisation flow.

